PR #12, opened 2015-11-19
global: multiple endpoints and pidstore integration

lnielsen:


examples/app.py
```diff
@@ -30,18 +30,87 @@
...
+            status=PIDStatus.REGISTERED)
+        Record.create({'title': 'Registered '}, id_=rec_uuid)
+
+        # Record 2 - Deleted PID with record
+        rec_uuid = uuid.uuid4()
+        pid = PersistentIdentifier.create(
+            'recid', '2', object_type='rec', object_uuid=rec_uuid,
+            status=PIDStatus.REGISTERED)
+        pid.delete()
+        Record.create({'title': 'Live '}, id_=rec_uuid)
```
nharraud: Wouldn't it be more clear to create the record first and then mark the pid as deleted?

invenio_records_rest/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def record_to_json_serializer(pid, record, code=200, headers=None):
+    """Build a json flask response using the given record.
+
+    :Returns: A flask response with json data.
+    :Returns Type: :py:class:`flask.Response`
+    """
+    # FIXME: use a formatter instead once it is implemented
+    self = record_self_link(pid, record, _external=True)
```
nharraud: Minor detail, it might be better to name this variable self_link as self has another meaning in python.

invenio_records_rest/views.py
```diff
@@ -0,0 +1,239 @@
...
+
+        The body should be a JSON object, which will fully replace the current
+        record metadata.
+
+        :param pid: Persistent identifier for record.
+        :param record: Record object.
+        :returns: The modified record.
+        """
+        # TODO: accept non json content (MARC21...)
+        data = request.get_json()
```
nharraud: request.get_json doc says: `By default this function will only load the json data if the mimetype is application/json`. Just to be sure I tested with `application/json-patch+json`. It fails.
What about doing this:
* if the content type is `application/json-patch+json` and call `request.get_json(froce=True)` then `record.patch(data)`.
* else call `request.get_json()`, if the result is None and content-type is `application/json` or not set at all, fail with 400, else fail with 415

nharraud:
Looks good to me. In the end I don't think that renaming `restful` files as `views` is an issue. Modules will most probably separate in directories the UI and the REST api anyway. Something like:
* package
  * views.py (UI)
  * v1
    * ext.py
    * serializer.py
    * views.py (API)
  * v2
    * ext.py
    * ....

Issue #14, opened 2015-11-20
views: integrate API authentication/access control


lnielsen:
- [ ] OAuth2 (with scopes)
- [ ] Access control

hachreak:
@lnielsen do you think we reached the result?

lnielsen:
- [ ]  OAuth2 (with scopes)
?

hachreak:
Should it be done by the permission factories? (see https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L392)

nharraud:
@lnielsen I am not sure how we can have oauth2 scopes in invenio-records-rest given that it is used in different modules. Shouldn't we have different scopes for the *published* records and the deposits?

I planned to check the scopes directly in the permission factories as recommended by @jirikuncar (at least that's what I understood).

lnielsen:
Yes, it should be done by permission factories, but perhaps we should provide some default scopes to rely on and which are easy to integrate?

Just thinking that right now, you can get an access token for deposit, and use that you read records, files, etc. which a user might not want a third-party to do.

For scopes I'm thinking something simple as either just ``records`` or alternatively two scopes ``records:read`` + ``records:write/admin/update``.

nharraud:
@lnielsen So you mean implementing optional scopes [like here](https://github.com/inveniosoftware/invenio-deposit/blob/180b50fa3b77cc3e17dfcc6ff10b7e59ca59f3ff/invenio_deposit/scopes.py) without creating any permission factory? If so, I like the idea. Having a `records:read` and `records:update` should be enough for most use cases and would enable to have scopes which can be reused in other modules if need be. They should just remain optional.

Note: `records:update` should not give access to the `delete` operation (which the admins can do). There should be another scope for delete, if it is needed.

lnielsen:
@nharraud possible the ``records:update/write`` scope could still be used to protect the ``delete`` operation, since in addition to the scope check, there's also a permission check which would fail unless you're admin…… the scopes is just to restrict what a third-party application can do with the delegated access.

nharraud:
@lnielsen I am not sure that as an admin I would give access to the "force" delete operation to any script. Most of the scripts would just get the right to fix/migrate records.

lnielsen:
@nharraud Yes, hence you shouldn't delegate ``record:update`` scope to the third-party app :-) However, IMHO you shouldn't even allow force delete to an admin user.

PR #18, opened 2015-12-03
views: access control

nharraud:
* NEW Adds customizable access control to record views. Allow
  configuring different permissions per endpoint. (reference #14)

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Examples instance directories
+examples/*_instance
--- a/invenio_records_rest/ext.py
+++ b/invenio_records_rest/ext.py
@@ -47,7 +47,7 @@ class _RecordRESTState(object):
         """Load default read permission factory."""
         if self._read_permission_factory is None:
             imp = self.app.config[
-                "RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY"
+                'RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY'
             ]
             self._read_permission_factory = import_string(imp) if imp else None
         return self._read_permission_factory
@@ -57,7 +57,7 @@ class _RecordRESTState(object):
         """Load default create permission factory."""
         if self._create_permission_factory is None:
             imp = self.app.config[
-                "RECORDS_REST_DEFAULT_CREATE_PERMISSION_FACTORY"
+                'RECORDS_REST_DEFAULT_CREATE_PERMISSION_FACTORY'
             ]
             self._create_permission_factory = import_string(imp) \
                 if imp else None
@@ -68,7 +68,7 @@ class _RecordRESTState(object):
         """Load default update permission factory."""
         if self._update_permission_factory is None:
             imp = self.app.config[
-                "RECORDS_REST_DEFAULT_UPDATE_PERMISSION_FACTORY"
+                'RECORDS_REST_DEFAULT_UPDATE_PERMISSION_FACTORY'
             ]
             self._update_permission_factory = import_string(imp) \
                 if imp else None
@@ -79,7 +79,7 @@ class _RecordRESTState(object):
         """Load default delete permission factory."""
         if self._delete_permission_factory is None:
             imp = self.app.config[
-                "RECORDS_REST_DEFAULT_DELETE_PERMISSION_FACTORY"
+                'RECORDS_REST_DEFAULT_DELETE_PERMISSION_FACTORY'
             ]
             self._delete_permission_factory = import_string(imp) \
                 if imp else None
@@ -99,7 +99,7 @@ class InvenioRecordsREST(object):
         self.init_config(app)
         # Register records API blueprints
         app.register_blueprint(
-            create_blueprint(app.config["RECORDS_REST_ENDPOINTS"])
+            create_blueprint(app.config['RECORDS_REST_ENDPOINTS'])
         )
         app.extensions['invenio-records-rest'] = _RecordRESTState(app)
 
@@ -107,7 +107,7 @@ class InvenioRecordsREST(object):
         """Initialize configuration."""
         # Set up API endpoints for records.
         app.config.setdefault(
-            "RECORDS_REST_ENDPOINTS",
+            'RECORDS_REST_ENDPOINTS',
             dict(
                 recid=dict(
                     pid_type='recid',
--- a/invenio_records_rest/ext.py
+++ b/invenio_records_rest/ext.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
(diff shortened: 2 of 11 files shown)


invenio_records_rest/ext.py
```diff
@@ -26,9 +26,66 @@
...
+
+    def __init__(self, app):
+        """Initialize state."""
+        self.app = app
+        self._read_permission_factory = None
+        self._create_permission_factory = None
+        self._update_permission_factory = None
+        self._delete_permission_factory = None
+
+    @property
```
jirikuncar: IMHO it can be `werkzeug.utils:cached_property`

invenio_records_rest/views.py
```diff
@@ -42,10 +42,29 @@
 from jsonpatch import JsonPatchException, JsonPointerException
 from sqlalchemy.exc import SQLAlchemyError
 from werkzeug.routing import BuildError
+from werkzeug.local import LocalProxy
+from werkzeug.utils import import_string
 
 from .serializers import record_to_json_serializer
 
 
+current_read_permission_factory = LocalProxy(
```
jirikuncar: isn't it enough to have just `current_records_rest`?

invenio_records_rest/views.py
```diff
@@ -194,6 +272,17 @@ def get(self, pid, record, **kwargs):
         :param record: Record object.
         :returns: The requested record.
         """
+        # Check permissions
```
jirikuncar: perfect candidate for a decorator

tests/test_invenio_records_rest.py
```diff
@@ -90,6 +90,7 @@ def control_num(data, cn=1):
 
 def test_valid_create(app, resolver):
     """Test VALID record creation request (POST .../records/)."""
+    InvenioRecordsREST(app)
```
jirikuncar: IMHO it should be part of `app` fixture

examples/permsapp.py
```diff
@@ -0,0 +1,170 @@
...
+Try to get record 1:
+
+    $ curl -XGET http://localhost:5000/records/1
+
+Login as admin@invenio-software.org (password 123456) and get record 1:
+
+    $ curl -XPOST http://localhost:5000/login -d 'email=admin%40invenio-software.org&password=123456' -c mycookie
+    $ curl -XGET http://localhost:5000/records/1 -b mycookie
+
+Login as forbidden@invenio-software.org (password 123456), hwo has not
```
lnielsen: hwo -> who

examples/permsapp.py
```diff
@@ -0,0 +1,170 @@
...
+InvenioREST(app)
+InvenioPIDStore(app)
+InvenioRecords(app)
+Menu(app)
+accounts = InvenioAccounts(app)
+app.register_blueprint(blueprint)
+InvenioAccess(app)
+InvenioRecordsREST(app)
+
+rec_uuid = '9107e6ef-fea4-4971-ae2f-934d2fdcaa30'
```
lnielsen: Probably we should use the same uuid as in records-ui:
``deadbeef-1234-5678-ba11-b100dc0ffee5``

invenio_records_rest/ext.py
```diff
@@ -44,10 +101,25 @@ def init_app(self, app):
...
+            "RECORDS_REST_DEFAULT_CREATE_PERMISSION_FACTORY",
+            "invenio_records.permissions:create_permission_factory")
+        app.config.setdefault(
+            "RECORDS_REST_DEFAULT_UPDATE_PERMISSION_FACTORY",
+            "invenio_records.permissions:update_permission_factory")
+        app.config.setdefault(
+            "RECORDS_REST_DEFAULT_DELETE_PERMISSION_FACTORY",
+            "invenio_records.permissions:delete_permission_factory")
+
+        app.config.setdefault("RECORDS_REST_LOGIN_ENDPOINT", "security.login")
```
lnielsen: Do we redirect in the rest API? shouldn't we just return a 401 error.

lnielsen: Yes, it does return 401 error, so I guess this one isn't needed?

nharraud:
@lnielsen @jirikuncar One thing you can see in the code but which I forgot to mention is that the tests and example use the UI login as we don't have oauth yet. As the search-ui will be also use cookie authentication to access this REST API this seemed a valid usage even if not the most common.

invenio_records_rest/ext.py
nharraud: Yes you are right, bad copy paste from invenio-records-ui...

jirikuncar:
@nharraud both OAuthClient and OAuth2Server should be ready on GitHub (not PyPI yet).

nharraud:
@jirikuncar Sorry I meant Application Token. There is no need for an OAuth authentication here. Using an application token added directly to the database would remove the dependency on UI login.

nharraud:
Coverage decreased because I added the delete permission factory methods even though there is no delete method implemented yet.

nharraud:
@jirikuncar can we release a new version of invenio-rest? This depends on the last commits.

nharraud:
@jirikuncar I removed the example with permissions, removed default permission factories and copied permissions.py from invenio_records for the tests.

invenio_records_rest/views.py
```diff
@@ -178,15 +266,52 @@ class RecordResource(ContentNegotiatedMethodView):
...
+
+        :param pid: Persistent identifier for record.
+        :param record: Record object.
+        """
+        self.check_etag(str(record.model.version_id))
+
+        try:
+            record.delete()
+            # mark all PIDs as DELETED
+            all_pids = PersistentIdentifier.query \
```
jirikuncar: IMHO you can use only one call of `.filter()` (or `db.and_(cond1, cond2)`).
```python
all_pids = PersistentIdentifier.query.filter(
    PersistentIdentifier.object_type == pid.object_type,
    PersistentIdentifier.object_uuid == pid.object_uuid
).all()
```

tests/test_invenio_records_rest.py
```diff
@@ -81,6 +84,18 @@ def create_record(data):
     return pid, record
 
 
+def delete_record(pid, record):
+    with db.session.begin_nested():
+        record.delete()
```
jirikuncar: We should consider hooking PIDStore marking as deleted to record signal.

--
cc @lnielsen

tests/permissions.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015, 2016 CERN.
```
jirikuncar: if it's a new file it should contain only current year

.gitignore
```diff
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Sqlite databases
+*.db
```
jirikuncar: the better solution would be to do proper cleaning after running tests

nharraud: In fact these are not for tests, which are cleaning properly. This is for the examples. The `app.py` example has its instance directory inside the `examples` directory for two reasons. The first one is that the directory was not created if it did not exist. The second reason is that every module's example is named `app,py` and thus has the same instance director. I had to clean it manually each time in order to be sure that the database was in a clean state.
Thus I create an instance directory for each example in the examples directory.

tests/test_invenio_records_rest.py
nharraud: @jirikuncar I am not sure doing the deletion automatically is the best way to do it. We would automate only part of the process. The pid creation is not automatic either. The records-rest API is creating both the pid and the record separately. It will also be complicated to detect the merging and mark the pid as *moved*.
Also the pidstore module does not depend on invenio-records (or only for the tests).

.gitignore
jirikuncar: Wouldn't `examples/instance/` do the job?

nharraud: I had multiple examples at that time, one without access control and one with it. I have removed it so that it does not depend on the deprecated invenio_records/permissions.py but I might add another example providing its own permission_factories later.

tests/test_invenio_records_rest.py
nharraud: @jirikuncar If it is only a concern on code duplication I will call the delete from the REST API instead. It will be cleaner that way.

invenio_records_rest/ext.py
```diff
@@ -26,9 +26,66 @@
...
+        self._create_permission_factory = None
+        self._update_permission_factory = None
+        self._delete_permission_factory = None
+
+    @cached_property
+    def read_permission_factory(self):
+        """Load default read permission factory."""
+        if self._read_permission_factory is None:
+            imp = self.app.config[
+                "RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY"
```
jirikuncar: It is not critical, but can you use single quotes for strings for consistency with rest of the code. They are randomly mixed through the code also in other files. Thanks

nharraud: Yes you are right. Can I fix that in the next pull request (search integration) so that I don't spend too much time rebasing?

jirikuncar: IMHO it's not so much pain to run search&replace. You won't have many conflicts when you rebase other branches.

nharraud: Ok. Is there a guideline for quotes in Invenio? I think we should also fix the cookie cutter as it is putting `"` in some places.

jirikuncar: Just use your judgement and be consistent.

tests/__init__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
jirikuncar: Please remove this file as its usage is discouraged with `py.test`. 

> **avoid “__init__.py” files in your test directories.** This way your tests can run easily against an installed version of mypkg, independently from the installed package if it contains the tests or not.

*source https://pytest.org/latest/goodpractises.html*

jirikuncar:
- [ ] Update copyright year in **all** files you have touched in this PR. Thanks

nharraud:
@jirikuncar Done. Only the `.gitignore` has not been updated as I didn't see copyrights in other projects either.

PR #26, opened 2016-02-01
global: default media type and tests refactoring

lnielsen:


invenio_records_rest/views.py
```diff
@@ -95,6 +96,7 @@ def create_url_rules(endpoint, list_route=None, item_route=None,
     :param search_type: Name of the search type used when searching records.
     :param record_serializers: serializers used for records.
     :param search_serializers: serializers used for search results.
+    :param default_media_type: default serializer for both records and search.
```
nharraud: `s/default serializer/default media-type used to choose the serializer/`. The serializers are different.

PR #30, opened 2016-02-05
views: max result window fix

lnielsen:


lnielsen:
Pending inveniosoftware/invenio-rest#23

PR #40, opened 2016-02-15
views: search result self link

lnielsen:
* Adds self link to search results page.


drjova:
LGTM :ship: ?

PR #41, opened 2016-02-16
views: query factory support

lnielsen:
* Add support for customizing query generation and thereby support
  multiple query parsers.

* Adds query factory for Elasticsearch Lucene syntax.


drjova:
LGTM

PR #52, opened 2016-03-29
serializers: datacite tests

lnielsen:
* Adds tests for DataCite serializer.

* Fixes issue with OAI-PMH serialization.

### DIFF
--- a/invenio_records_rest/serializers/datacite.py
+++ b/invenio_records_rest/serializers/datacite.py
@@ -71,7 +71,7 @@ class DataCite31Serializer(MarshmallowSerializer):
 
     def serialize_oaipmh(self, pid, record):
         """Serialize a single record for OAI-PMH."""
-        obj = self.transform_record(pid, record) \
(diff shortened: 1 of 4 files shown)


nharraud:
@lnielsen looks good for me. Can I merge or are you still working on it?

Maybe we should create an issue for removing fields from the records' metadata returned by search. I don't think we have such a mechanism yet. It will be needed to remove sensitive information before returning them to users, whatever the serializer.

lnielsen:
It's ready from my site. Thanks for the review.

lnielsen:
Removing sensitive data
- [This](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/indexer.py#L43-L44) ensures that sensitive data cannot be searched on (when using bare ES syntax).
- [This](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/__init__.py#L65) serialiser removes stuff which is not for the public eye.

PR #57, opened 2016-04-04
views: suggest endpoints addition

### LINKED ISSUE #54: views: completer suggester endpoint
Add a new ``../_suggest`` endpoint to Records-REST which will use https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html in the backend.

Use case: Auto-completion in deposit forms and search fields.

hachreak:
* Adds suggest endpoints. (closes #54)

### DIFF
--- a/invenio_records_rest/config.py
+++ b/invenio_records_rest/config.py
@@ -54,6 +54,19 @@ RECORDS_REST_ENDPOINTS = dict(
         item_route='/records/<pid_value>',
         default_media_type='application/json',
         max_result_window=10000,
+        suggestions=[
+            dict(
+                route='/records/_suggestions',
+                opts=dict(
+                    completion=dict(
+                        field='suggest_title',
+                        fuzzy=dict(
+                            fuzziness=2
+                        )
+                    )
+                )
+            )
+        ]
     ),
 )
 
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -26,7 +26,9 @@
 
 from __future__ import absolute_import, print_function
 
+import json
 import uuid
+from copy import deepcopy
 from functools import partial, wraps
 
 from flask import Blueprint, abort, current_app, jsonify, make_response, \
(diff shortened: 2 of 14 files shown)


jirikuncar:
We should keep naming consistent: suggestions vs. completions (/api/records/suggest vs /api/records/complete)?

invenio_records_rest/views.py
```diff
@@ -277,6 +299,73 @@ def need_record_permission_decorator(self, record, *args, **kwargs):
...
+
+        query = {
+            self.completion_opts['field']: {
+                'text': text,
+                'completion': self.completion_opts
+            }
+        }
+
+        # Execute search
+        search_result = current_search_client.suggest(
```
lnielsen: We should make sure we don't suggest restricted records.

hachreak: Can we use the `Context` to limit the query?
> each context value generates a new sub-set of documents which can be queried by the completion suggester

see https://www.elastic.co/guide/en/elasticsearch/reference/current/suggester-context.html

lnielsen: Hmmm....actually this is not needed to begin with since we're only going to auto-complete from fully public vocabularies (perhaps we can post-pone it until we have to deal with search auto-completion).

hachreak:
ping @lnielsen the suggester is ready to use. 
I also modified the example app to use it.

invenio_records_rest/config.py
```diff
@@ -54,6 +54,19 @@ def _(x):
         item_route='/records/<pid_value>',
         default_media_type='application/json',
         max_result_window=10000,
+        suggestions=[
```
jirikuncar: why is it a list?

hachreak: I tried to be more generic as possible. In this case each element of the array is a suggestion.
If e.g you need to define different suggestions for different fields.

hachreak:
ping @lnielsen

lnielsen:
Merged via #63

PR #62, opened 2016-04-13
global: usage of elasticsearch-dsl

jirikuncar:
Signed-off-by: Jiri Kuncar <jiri.kuncar@cern.ch>


invenio_records_rest/views.py
```diff
@@ -160,7 +159,7 @@ def create_url_rules(endpoint, list_route=None, item_route=None,
                                        with_deleted=True))
 
     list_view = RecordsListResource.as_view(
-        RecordsListResource.view_name.format(endpoint),
+        RecordsListResource.view_name.format(pid_type),
```
lnielsen: reminder revert

tests/access_records.py
```diff
@@ -115,24 +116,15 @@ def index_record_modification(sender, changes):
...
-                NotOp(KeywordOp(Keyword('_access.exclude_users'),
-                                DoubleQuotedValue(current_user.id)))
-            ),
-            query.query
-        )
+        return Q('match', **{'_access.public': 1})
+    return Q(
+        'match', **{'_access.include_users': current_user.id}
+    ) & ~Q(
+        'match', **{'_access.exlude_users': current_user.id}
```
jmartinm: _access.exlude_users -> _access.exclude_users

jmartinm:
Thanks for the changes.

So for current users of `invenio-records-rest` the changes in the overlay would be to use `search_factory_imp` instead of `query_factory_imp`?  Then this PR adds the possibility of filtering records (for access control for example).

Any other changes that could affect our overlays?

jirikuncar:
@jmartinm we are going to refactor invenio-search. There will be changing coming. Stay tuned

jmartinm:
> @jmartinm we are going to refactor invenio-search. There will be changing coming. Stay tuned

Improvements are welcome, as long as it does not break our code too badly ;)

jirikuncar:
@jmartinm they will. We are still in alpha mode.

lnielsen:
@jmartinm Changes are necessary upfront in order avoid a future refactoring *after* final release (which IMHO is worse than taking the pain now)

jmartinm:
> @jmartinm Changes are necessary upfront in order avoid a future refactoring after final release (which IMHO is worse than taking the pain now)

Yes, no worries. I am sure the changes will be for the best 👍

invenio_records_rest/query.py
```diff
@@ -27,60 +27,38 @@
...
 
 from flask import current_app, request
-from invenio_search import Query
 
 from .errors import InvalidQueryRESTError
 
 
-def default_query_factory(index, page, size):
-    """Parse and slice query using Invenio-Query-Parser.
+def default_query_factory(self, search, query_parser=None):
```
lnielsen: Should we rename to ``default_search_factory``?

PR #71, opened 2016-05-02
global: PID resolver as URL converter

jirikuncar:
@lnielsen can you please review it? We can now use `pid, record = request.view_args['pid_value']` anywhere.

I would like to also remove `@pass_record` decorator.

--
- [ ] release Invenio-Base
- [ ] update Invenio-Deposit and overlay configurations (ping @inveniosoftware/invenio-service-managers)


lnielsen:
LGTM - we should apply same pattern in Records-UI

PR #77, opened 2016-05-04
global: lazy PID value evaluation

jirikuncar:
* Makes evaluation of URL argument of `pid` type lazy.

Signed-off-by: Jiri Kuncar <jiri.kuncar@cern.ch>

--
ups ...


nharraud:
This fixes my issue on B2Share. 👍

PR #80, opened 2016-05-10
views: addition of links in response headers

### LINKED ISSUE #70: views: links not included in headers
self, next, and prev links are not included in the HTTP headers.

https://tools.ietf.org/html/rfc5988

omelkonian:
* FIX Fixes missing links between URIs in response headers. These
  links were included in the JSON object, but not in the HTTP headers,
  as suggested on https://tools.ietf.org/html/rfc5988.  (closes #70)

* Increases test coverage.

### DIFF
--- a/invenio_records_rest/serializers/response.py
+++ b/invenio_records_rest/serializers/response.py
@@ -47,7 +47,12 @@ def record_responsify(serializer, mimetype):
         response.last_modified = record.updated
         if headers is not None:
             response.headers.extend(headers)
+
+        if links_factory is not None:
+            add_link_header(response, links_factory(pid))
+
         return response
+
     return view
 
 
@@ -67,5 +72,23 @@ def search_responsify(serializer, mimetype):
         response.status_code = code
         if headers is not None:
             response.headers.extend(headers)
+
+        if links is not None:
+            add_link_header(response, links)
+
         return response
+
     return view
+
+
+def add_link_header(response, links):
+    """Add a Link HTTP header to a REST response.
+
+    :param response: REST response instance
+    :param links: Dictionary of links
+    """
+    if links is not None:
+        response.headers.extend({
+            'Link': ', '.join([
+                  '<{0}>; rel="{1}"'.format(l, r) for r, l in links.items()])
+        })
--- a/tests/test_views_search.py
+++ b/tests/test_views_search.py
@@ -26,6 +26,8 @@
 
 from __future__ import absolute_import, print_function
 
+import re
+
 import pytest
 from helpers import assert_hits_len, get_json, parse_url, to_relative_url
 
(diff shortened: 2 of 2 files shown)


invenio_records_rest/views.py
```diff
@@ -403,13 +403,22 @@ def get(self, **kwargs):
...
 
-        return self.make_response(
+        response = self.make_response(
             pid_fetcher=self.pid_fetcher,
             search_result=search_result.to_dict(),
             links=links,
             item_links_factory=self.item_links_factory,
         )
 
+        # Add Link HTTP header
```
lnielsen: This fits better in [record_responsify](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/serializers/response.py#L48-L49) and [search_responsify](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/serializers/response.py#L68-L69).

invenio_records_rest/serializers/response.py
```diff
@@ -47,6 +47,15 @@ def view(pid, record, code=200, headers=None, links_factory=None):
         response.last_modified = record.updated
         if headers is not None:
             response.headers.extend(headers)
+
```
lnielsen:
```python
def add_link_header(response, links):
    if links is not None:
        response.headers.extend({
            'Link': ', '.join([
                '<{0}>; rel="{1}"'.format(l, r) for r, l in links.items()])
        })
```

...

```
if links_factory:
    add_link_header(response, links_factory(pid))
```

invenio_records_rest/serializers/response.py
```diff
@@ -67,5 +76,13 @@ def view(pid_fetcher, search_result, code=200, headers=None, links=None,
         response.status_code = code
         if headers is not None:
             response.headers.extend(headers)
+
+        # Add Link HTTP header
+        if links is not None:
```
lnielsen:
```python
if links:
    add_link_header(response, links)
```

tests/test_views_search.py
```diff
@@ -93,6 +95,50 @@ def test_pagination(app, indexed_records, search_url):
...
+        parsed_url = parse_url(next_url)
+        assert parsed_url['qs']['size'] == ['1']
+        assert parsed_url['qs']['page'] == ['2']
+
+        # Access next URL
+        res = client.get(to_relative_url(next_url))
+        assert_hits_len(res, 1)
+        links = {
+            k: v for (k, v) in
+            map(lambda s: re.findall(r'<(.*)>; rel="(.*)"', s)[0][::-1],
```
lnielsen: This line is the same as L106-109 so can be put in it's own function (a la ``parse_link_header``).

egabancho:
Just need to add ` Signed-off-by:` (`git commit -s`) otherwise LGTM, @lnielsen?

PR #81, opened 2016-05-10
views: read permission on get all records

hachreak:

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -368,6 +368,7 @@ class RecordsListResource(ContentNegotiatedMethodView):
             current_records_rest.loaders
         self.record_class = record_class or Record
 
+    @need_record_permission('read_permission_factory')
     def get(self, **kwargs):
         """Search records.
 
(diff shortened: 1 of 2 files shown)


invenio_records_rest/views.py
```diff
@@ -368,6 +368,7 @@ def __init__(self, resolver=None, minter_name=None, pid_type=None,
             current_records_rest.loaders
         self.record_class = record_class or Record
 
+    @need_record_permission('read_permission_factory')
```
jirikuncar: `@require_api_auth(allow_anonymous=True)` - see https://github.com/inveniosoftware/invenio-oauth2server/pull/41

hachreak: I think depends from the kind of record.
In the case of zenodo deposit, I need to block anonymous users.
See quickstart: https://zenodo.org/dev#collapse-list0

lnielsen:
Add a ``search_factory_imp`` to DEPOSIT REST config. Wrap invenio_records_rest.query.es_search_factory and add access restriction.

PR #83, opened 2016-05-18
facets: new filter for ranges

nikofil:
* NEW Implements a filter that filters a field based on a range of
 values. This is used in the year filter facet for invenio-search-ui.

### DIFF
--- a/invenio_records_rest/facets.py
+++ b/invenio_records_rest/facets.py
@@ -27,7 +27,9 @@
 from __future__ import absolute_import, print_function
 
 from elasticsearch_dsl import Q
+from elasticsearch_dsl.query import Range
 from flask import current_app, request
+from invenio_rest.errors import FieldError, RESTValidationError
 from werkzeug.datastructures import MultiDict
 
 
@@ -35,6 +37,45 @@ def terms_filter(field):
     """Create a term filter."""
     def inner(values):
         return Q('terms', **{field: values})
+    return inner
+
+
+def range_filter(field, start_date_math=None, end_date_math=None, **kwargs):
+    """Create a range filter."""
+    def inner(values):
+        if len(values) != 1 or values[0].count('--') != 1 or values[0] == '--':
+            raise RESTValidationError(
+                errors=[FieldError(field, 'Invalid range format.')])
+
+        range_ends = values[0].split('--')
+        range_args = dict()
+
+        ineq_opers = [{'strict': 'gt', 'nonstrict': 'gte'},
+                      {'strict': 'lt', 'nonstrict': 'lte'}]
+        date_maths = [start_date_math, end_date_math]
+
+        # Add the proper values to the dict
+        for (range_end, strict, opers,
+             date_math) in zip(range_ends, ['>', '<'], ineq_opers, date_maths):
+
+            if range_end != '':
+                # If first char is '>' for start or '<' for end
+                if range_end[0] == strict:
+                    dict_key = opers['strict']
+                    range_end = range_end[1:]
+                else:
+                    dict_key = opers['nonstrict']
+
+                if date_math:
+                    range_end = '{0}||{1}'.format(range_end, date_math)
+
+                range_args[dict_key] = range_end
+
+        args = kwargs.copy()
+        args.update(range_args)
+
+        return Range(**{field: args})
+
     return inner
 
 
--- a/tests/test_facets.py
+++ b/tests/test_facets.py
@@ -27,19 +27,38 @@
 
 from __future__ import absolute_import, print_function
 
+import pytest
 from elasticsearch_dsl import Search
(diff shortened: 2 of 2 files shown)


invenio_records_rest/facets.py
```diff
@@ -38,6 +38,27 @@ def inner(values):
     return inner
 
 
+def range_filter(field, **kwargs):
+    """Create a range filter."""
+    def inner(values):
+        range_ends = [int(val) for val in values[0].split(',')]
```
jirikuncar: - [ ] We should be able to support not only `int`.
- [ ] Why using `,` as separator for ranges?
- [ ] Have you considered using Elasticsearch-DSL `Range` instead?

invenio_records_rest/facets.py
```diff
@@ -38,6 +39,42 @@ def inner(values):
     return inner
 
 
+def range_filter(field, **kwargs):
```
lnielsen:
```python
def range_filter(field, start_date_math=None, end_date_math=None, **kwargs):
    """Create a range filter."""
    def inner(values):
        start, end = values[0].split('--')

        if start_date_math:
            start = '{0}||{1}'.format(start, start_date_math)

        if end_date_math:
            end = '{0}||{1}'.format(end, end_date_math)

        kwargs.update({
            'gte': start,
            'lte': end
        })

        return Range(**{field: params})

    return inner
```

invenio_records_rest/facets.py
```diff
@@ -38,6 +39,42 @@ def inner(values):
...
+    if 'start_date_math' in kwargs:
+        start_date_math = '||' + args['start_date_math']
+        del args['start_date_math']
+
+    if 'end_date_math' in kwargs:
+        end_date_math = '||' + args['end_date_math']
+        del args['end_date_math']
+
+    def inner(values):
+        [start, end] = values[0].split('--')
```
lnielsen: What happens on bad input data - e.g. ``1940`` or ``1940-``

nikofil: If no '--' is found, the terms_filter will be now used instead, as I believe 1940- may be a valid value in some context.

invenio_records_rest/facets.py
jirikuncar: `s/return Range(**{field: params})/return Range(**{field: kwargs})/`?

lnielsen:
The ES [range query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html) falls in two types:

- TermRangeQuery (strings)
- NumericRangeQuery (dates + numbers)

Both types needs one or more of the following parameters: ``gt``, ``gte``, ``lt``, ``lte``

The REST API can be queried like this:

- ``/api/records/?test=<value1>&test=<value2>``

With will result in *one* call to the ``inner`` method with ``values`` being a list ``[<value 1>, <value 2>]``

Issues:
- Handling multiple input values -> should produce multiple range queries or perhaps fail.
- Parsing and validating of one input value, and mapping this into one or more of the parameters (``gt``, ``gte``, ``lt``, ``lte``). 

Regarding the parsing we could do something along the lines of:
- ``1940--2009`` -> ``gte=1940, lte=2009``
- ``--2009`` -> ``lte=2009``
- ``1940--`` -> ``gte=1940``
- ``1940`` -> ``gte=1940``
- ``>1940--<2009`` -> ``gt=1940, lt=2009``
- ``--<2009`` -> ``lt=2009``
- ``>1940--`` -> ``gt=1940``
- ``a--m`` -> ``gte=a, lte=m``
- ...

In case the parsing fails, I think we should raise a ``invenio_rest.errors.RESTValidationError``:

```python
raise RESTValidationError(errors=[FieldError('test', 'Invalid range format.')])
```

This will result in a 400 HTTP error, with a message staying that the format is invalid.

nikofil:
In which case would the parsing fail? Only case I can think of would be just `--`.
Besides that, I don't see how multiple values would fit into the range query, unless we did something like `term=gt2016&term=lte2020` instead of `2016--<2020`. Of course, I have little experience with ElasticSearch, so any suggestions are welcome.

tests/test_facets.py
```diff
@@ -42,6 +45,26 @@ def test_terms_filter():
...
+    assert f(['1821--1940']) == Range(test={
+        'gte': '1821||startmath', 'lte': '1940||endmath',
+    })
+    assert f(['>1821--']) == Range(test={'gt': '1821||startmath'})
+    assert f(['1821--<1940']) == Range(test={'gte': '1821||startmath',
+                                             'lt': '1940||endmath'})
+
+    exception_raised = False
+    try:
+        f(['2016'])
```
lnielsen:
```python
assert pytest.raises(RESTValidationError, f, ['2016'])
```

PR #86, opened 2016-06-01
WIP: REST endpoints refactoring

nharraud:


nharraud:
@jirikuncar @lnielsen this is still WIP but I would like to know what you think of this.

The endpoint configuration is parsed by a Marshmallow schema which creates a config object. This config handles endpoint vs default values, thus the defaults can be overriden in deposit by replacing the schema class.

The schema is used by the extension's `init_app` and the result is given to the `create_blueprint` function. This enables to change the configuration schema and result classes and still use `create_blueprint`.

The config is given to each View class. It is then added to the application context on each REST request thanks to the decorator `with_endpoint_config`. The endpoint_config can be retrieved from anywhere by using `config_loader.current_endpoint_config`.

There is one thing I don't know, it is if we use the endpoint_config for everything. For example I started removing the `links_factory` from the serializers arguments. The base serializer can directly retrieve the `links_factory` by accessing `config_loader.current_endpoint_config`. Is this something we want?

nharraud:
feedback from @lnielsen : use request context instead of application context.

jirikuncar:
> There is one thing I don't know, it is if we use the endpoint_config for everything. For example I started removing the `links_factory` from the serializers arguments. The base serializer can directly retrieve the `links_factory` by accessing `config_loader.current_endpoint_config`. Is this something we want?

I would keep the way how the `links_factory` is passed to `ContentNegotiatedView` untouched for now.

invenio_records_rest/config_loader.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if search_class_kwargs:
+            data['search_class'] = partial(data['search_class'],
+                                           **search_class_kwargs)
+
+        data['resolver'] = Resolver(
+            pid_type=data['pid_type'],
+            object_type='rec',
+            getter=partial(data['record_class'].get_record, with_deleted=True))
+        return EndpointConfig(**data)
```
jirikuncar: Do we need to wrap the data to another class? Can't we have all the "cached properties" as special fields on schema?

nharraud: You mean as properties, additional methods or dict fields containing functions?
We can't have them as values as there are some information inaccessible at the time of the config deserialization (ex: `current_pidstore`).

nharraud: NOTE after discussion: We will use the resulting dict instead of a custom class. Some fields will use LocalProxy to resolve their value later.

lnielsen:
Closing PR as it's been stale for a long time.

PR #97, opened 2016-07-04
errors: more specific error messages

### LINKED ISSUE #32: errors: more specific error messages
Records-REST uses Flask's ``abort`` method to signal errors. E.g.:

```python
abort(415)
```

Invenio-REST will convert this into a response to the client which looks like this:

```http
HTTP/1.0 406 NOT ACCEPTABLE
Content-Type: application/json
Server: Werkzeug/0.11.3 Python/3.5.0
Date: Mon, 08 Feb 2016 13:35:48 GMT

{
  "message": "The resource identified by the request <...>.",
  "status": 406
}
```

The message is a generic error message which doesn't help client to figure out what they did wrong. Instead we should as much as possible provide proper error messages as to why a specific request went wrong.

This task is to review all ``abort`` calls, to check if a better error message can be provided. ``MaxResultWindowRESTError`` provides an example of how better error message can be provided.

omelkonian:
* Provides better error messages, instead of just using Flask's `abort`.
  (closes #32)

Signed-off-by: Orestis Melkonian <melkon.or@gmail.com>

### DIFF
--- a/invenio_records_rest/errors.py
+++ b/invenio_records_rest/errors.py
@@ -26,18 +26,154 @@
 
 from __future__ import absolute_import, print_function
 
+from flask import request
 from invenio_rest.errors import RESTException
 
 
+#
+# Search
+#
 class MaxResultWindowRESTError(RESTException):
     """Maximum number of results passed."""
 
     code = 400
-    description = "Maximum number of results have been reached."
+    description = 'Maximum number of results have been reached.'
 
 
+#
+# Query
+#
 class InvalidQueryRESTError(RESTException):
     """Invalid query syntax."""
 
     code = 400
-    description = "Invalid query syntax."
+    description = 'Invalid query syntax.'
+
+
+#
+# CiteProc
+#
+class StyleNotFoundRESTError(RESTException):
+    """No such style."""
+
+    code = 400
+
+    def __init__(self, style=None, **kwargs):
+        """Initialize exception."""
+        super(RESTException, self).__init__(**kwargs)
+        self.description = 'Style{0}could not be found.'.format(
+            ' "{0}" '.format(style) if style else ' ')
+
+
+#
+# PID
+#
+class PIDDoesNotExistRESTError(RESTException):
+    """Non-existent PID."""
+
+    code = 404
+    description = 'PID does not exist.'
+
+
+class PIDUnregisteredRESTError(RESTException):
+    """Unregistered PID."""
+
+    code = 404
+    description = 'PID is not registered.'
+
+
+class PIDDeletedRESTError(RESTException):
+    """Deleted PID."""
+
+    code = 410
+    description = 'PID has been deleted.'
+
+
+class PIDMissingObjectRESTError(RESTException):
+    """PID missing object."""
+
+    code = 500
+
+    def __init__(self, pid, **kwargs):
+        """Initialize exception."""
+        super(RESTException, self).__init__(**kwargs)
+        self.description = 'No object assigned to {0}.'.format(pid)
+
+
+class PIDRedirectedRESTError(RESTException):
+    """Invalid redirect for destination."""
+
+    code = 500
+
+    def __init__(self, pid_type=None, **kwargs):
+        """Initialize exception."""
+        super(RESTException, self).__init__(**kwargs)
+        self.description = (
+            'Invalid redirect - pid_type{0}endpoint missing.'.format(
+                ' "{0}" '.format(pid_type) if pid_type else ' ')
+        )
+
+
+#
+# Views
+#
+class PIDResolveRESTError(RESTException):
+    """Invalid PID."""
+
+    code = 500
+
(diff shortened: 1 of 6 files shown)


invenio_records_rest/errors.py
```diff
@@ -29,15 +29,169 @@
...
+#
+# Search
+#
 class MaxResultWindowRESTError(RESTException):
     """Maximum number of results passed."""
 
     code = 400
-    description = "Maximum number of results have been reached."
+
+    def __init__(self, max):
```
jirikuncar: - please choose variable name that does not conflict with build-in function name
- provide a default value

invenio_records_rest/errors.py
```diff
@@ -29,15 +29,169 @@
...
+#
+class StyleNotFoundRESTError(RESTException):
+    """No such style."""
+
+    code = 400
+
+    def __init__(self, style):
+        """Initialize exception."""
+        self.description = \
+            '[CiteProc]: Style \'{0}\'could not be found.'.format(style)
```
jirikuncar: - Can you remove the "[prefixes]" in error messages?
- I would simplify the message to: `Style {0} could not be found.`

invenio_records_rest/errors.py
```diff
@@ -29,15 +29,169 @@
...
+
+
+class PIDRedirectedRESTError(RESTException):
+    """Invalid redirect for destination."""
+
+    code = 500
+
+    def __init__(self, pid_type):
+        """Initialize exception."""
+        self.description = 'Invalid redirect - pid_type \'{0}\' ' \
```
jirikuncar: Can you use consistent style for breaking long strings?

```
self.description = ("This is very long description "  # no backslash \
                    "that can not fit on one line.")

File: invenio_records_rest/serializers/citeproc.py [hunk 4]
```diff
@@ -37,6 +37,8 @@
 from webargs import fields
 from webargs.flaskparser import FlaskParser
 
+from invenio_records_rest.errors import StyleNotFoundRESTError
```
jirikuncar: we use relative imports for modules from current package:
```
from ..errors import StyleNotFoundRESTError
```

invenio_records_rest/utils.py
```diff
@@ -35,6 +35,10 @@
 from werkzeug.routing import BaseConverter, BuildError, PathConverter
 from werkzeug.utils import cached_property, import_string
 
+from invenio_records_rest.errors import PIDDeletedRESTError, \
```
jirikuncar: see https://github.com/inveniosoftware/invenio-records-rest/pull/97/files#r69448616

invenio_records_rest/errors.py
```diff
@@ -29,15 +29,169 @@
...
+        """Initialize exception."""
+        self.description = 'PID #{0} could not be resolved.'.format(pid)
+
+
+class CreateUnsupportedMediaRESTError(RESTException):
+    """Creating record with unsupported media type."""
+
+    code = 415
+
+    def __init__(self, type):
```
jirikuncar:
```
def __init__(self, content_type=None):
    """..."""
    content_type = content_type or request.content_type
```

invenio_records_rest/errors.py
```diff
@@ -29,15 +29,169 @@
...
+        """Initialize exception."""
+        self.description = \
+            '[Put record]: Unsupported media type \'{0}\'.'.format(type)
+
+
+class PutInvalidDataRESTError(RESTException):
+    """Invalid request body when replacing record."""
+
+    code = 400
+
```
jirikuncar: remove this empty line to follow formating style in the current file

omelkonian: But doesn't it break the PEP8 rule?

invenio_records_rest/errors.py
tiborsimko: Besides, please don't use this:

```
      'this is \'something\` isn't it?' 
```
     
but rather one of these:

```
      'this is "something" isn't it?' 
      "this is 'something' isn't it?" 
```

see [PEP-8](https://www.python.org/dev/peps/pep-0008/):

> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.

invenio_records_rest/errors.py
omelkonian: What default value should I put? I assume `None`?

jirikuncar: IMHO you don't even need to show the max and just keep the original description.

omelkonian: OK, should I put default value on all constructors?

jirikuncar: In most of the places it makes sense. Try to think about proper inheritance for each exception.

PR #103, opened 2016-07-19
examples: addition of requirements.txt

jacquerie:
In a similar spirit to https://github.com/inveniosoftware/invenio-records-ui/pull/45, here's some documentation updates for the example and the addition of a `requirements.txt`, as suggested in the [informal RFC](https://gitter.im/inveniosoftware/invenio?at=578c2b2d196179690efc8cc9).

### DIFF
--- a/examples/requirements.txt
+++ b/examples/requirements.txt
@@ -1,0 +1,3 @@
+invenio-db>=1.0.0a10
+invenio-indexer>=1.0.0a6
+invenio-records-rest[memento]>=1.0.0a15
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -37,6 +37,7 @@ recursive-include docs *.rst
 recursive-include docs Makefile
 recursive-include examples *.json
 recursive-include examples *.py
+recursive-include examples *.txt
 recursive-include invenio_records_rest *.mo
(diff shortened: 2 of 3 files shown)


examples/requirements.txt
```diff
@@ -0,0 +1,50 @@
+amqp==1.4.9
```
jirikuncar: @jacquerie please specify only top level requirements and use `>=X.Y,X.Y+1` if possible.

jacquerie: Ok. Won't happen straight away, I'm busy with another thing right now.

jacquerie: It happened. I'm not sure if with `>=X.Y` you meant to avoid specifying the patch level, but in any case this is currently impossible:
```shell
% pip install "invenio-db>=1.0"
Collecting invenio-db>=1.0
  Could not find a version that satisfies the requirement invenio-db>=1.0 (from versions: 1.0.0a1, 1.0.0a2, 1.0.0a3, 1.0.0a4, 1.0.0a5, 1.0.0a6, 1.0.0a7, 1.0.0a8, 1.0.0a9, 1.0.0a10)
No matching distribution found for invenio-db>=1.0
```

PR #107, opened 2016-08-31
views: error handlers move

lnielsen:
* Moves creation of errors handlers into separate function which can
  be used by e.g. Invenio-Deposit to install the same error handlers.

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -70,22 +70,8 @@ elasticsearch_error_handlers = {
 """List of handlers for ElasticSearch errors."""
 
 
-def create_blueprint(endpoints):
-    """Create Invenio-Records-REST blueprint.
-
-    :params: Dictionary representing the endpoints configuration.
-    :returns: Configured blueprint.
-    """
-    blueprint = Blueprint(
-        'invenio_records_rest',
-        __name__,
-        url_prefix='',
-    )
-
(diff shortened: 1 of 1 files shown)


invenio_records_rest/views.py
```diff
@@ -70,22 +70,8 @@ def elasticsearch_query_parsing_exception_handler(error):
...
-        'invenio_records_rest',
-        __name__,
-        url_prefix='',
-    )
-
-    for endpoint, options in (endpoints or {}).items():
-        for rule in create_url_rules(endpoint, **options):
-            blueprint.add_url_rule(**rule)
-
+def create_error_handlers(blueprint):
```
jirikuncar: Shall we call it from `create_blueprint` function?

lnielsen: Hehe :-) Probably a good idea ;-)

invenio_records_rest/views.py
```diff
@@ -110,6 +96,25 @@ def first(function, iterable, default=None):
...
+        'invenio_records_rest',
+        __name__,
+        url_prefix='',
+    )
+
+    for endpoint, options in (endpoints or {}).items():
+        for rule in create_url_rules(endpoint, **options):
+            blueprint.add_url_rule(**rule)
+
+    return create_error_handlers(blueprint)
```
jirikuncar: 👍

PR #109, opened 2016-08-31
views: use endpoint pid_type instead of key for blueprints

### LINKED ISSUE #101: RFC URL endpoint building
There is a specific reason why to construct the url endpoint is used the `key` of `RECORDS_REST_ENDPOINTS` and sometimes the `pid_type`?
See:
 - construction [with key](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L227) and [here](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L247), [here](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L265) and [here](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L277) `RecordsListOptionsResource.view_name.format(endpoint),`
 - construction [with pid_type](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L450)
 - construction [with pid_type](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L513) from the record. `endpoint = '.{0}_item'.format(pid.pid_type)`

Can we simplify and use only one way?
Maybe the best solution is to use the `pid_type` everywhere. WDYT? There are some contraindications?

-----------------------------------------------------------
ping @jirikuncar @lnielsen @tiborsimko

nikofil:
* Changes views.py to use pid_type in the endpoint map instead of the key for
  the name of the endpoint. (closes #101)

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -82,8 +82,8 @@ def create_blueprint(endpoints):
         url_prefix='',
     )
 
-    for endpoint, options in (endpoints or {}).items():
-        for rule in create_url_rules(endpoint, **options):
+    for options in (endpoints or {}).values():
+        for rule in create_url_rules(options['pid_type'], **options):
             blueprint.add_url_rule(**rule)
 
     # catch record validation errors


lnielsen:
I'm quite sure this will create conflicts if you have multiple endpoints for the same pid type.

```
RECODRS_REST_ENDPOINTS = {
  'recid' : {'pid_type': 'recid', ...},
  'recid_donwloads' : {'pid_type': 'recid', ...},
}
```

This pattern is very common in Records-UI. I'm not sure we're using in for Records-REST though.

hachreak:
I think we have conflicts in any case.
See for [example](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L450) where the `pid_type` is used to build the endpoint.

lnielsen:
@hachreak That I would classify as a bug :-)

The only case where we use the pid_type to generate the endpoint with is when we have to make a redirect. https://github.com/inveniosoftware/invenio-records-ui/blob/master/invenio_records_ui/views.py#L171-L173

Here the assumption is that you must define at one endpoint with the same name as a pid type - otherwise the redirects won't work.

hachreak:
Ah.. ok.. XD
So, in this case, maybe, does it means that we have at least 4 bugs? :) [1](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L513), [2](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L450), [3](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/links.py#L36), [4](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/utils.py#L132)

lnielsen:
1, 2 are bugs, 3 i don't think you can fix, 4 is not a bug

jirikuncar:
@nikofil @lnielsen @hachreak for 1,2 we should probably use [`self.__name__`](https://github.com/pallets/flask/blob/master/flask/views.py#L98) instead.

kaplun:
cc @kaplun @jacquerie @jmartinm 
This PR is needed before inspirehep/inspire-next#1613

kaplun:
Actually I agree with @lnielsen in that the proposed solution wouldn't work. What about the otherway round and not using the `pid_type` at all for building routes?

jirikuncar:
> The only case where we use the pid_type to generate the endpoint with is when we have to make a redirect. https://github.com/inveniosoftware/invenio-records-ui/blob/master/invenio_records_ui/views.py#L171-L173

@kaplun how do you propose to solve the redirection?

kaplun:
What about introducing a separate new config variable to be filled up with the explicit map `pid_type`->`endpoint` to be used in these situations?

kaplun:
Alternatively, since we are already in the `view`, and since PIDs can redirect to each other only within the same `pid_type` what about using the current endpoint of the view and just change the PID value?

kaplun:
@jirikuncar @lnielsen what do you think about #119 as an alternative fix to this issue?

jirikuncar:
Closed via #119.

PR #111, opened 2016-09-02
tests: addition of test_example_app.py

### LINKED ISSUE #108: tests: addition of test_example_app.py
The package contains `examples/app.py` example application, but there is no test
for it in `tests/test_example_app.py`.  It should be added.  See existing examples:

- https://github.com/inveniosoftware/invenio-records/blob/master/tests/test_example_app.py
- https://github.com/inveniosoftware/invenio-marc21/blob/master/tests/test_example_app.py

hachreak:
* Adds test_example_app.py. (closes #108)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,6 @@ env:
 
 python:
   - "2.7"
-  - "3.3"
-  - "3.4"
   - "3.5"
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ cache:
 
 services:
   - postgresql
+  - rabbitmq
   - redis
 
 env:
--- a/examples/app-setup.sh
+++ b/examples/app-setup.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+# quit on errors:
+set -o errexit
+
+# quit on unbound symbols:
+set -o nounset
+
+export FLASK_APP=app.py
+
+# Install specific dependencies
+pip install -r requirements.txt
+
+# Create the database
+flask db init
+flask db create
+
+# Initialize indices
+flask index init
+flask index queue init
--- a/examples/app.py
+++ b/examples/app.py
@@ -27,18 +27,20 @@
 
 For this example the access control is disabled.
 
-Run example development server:
+Install requirements:
 
 .. code-block:: console
 
+    $ pip install -e .[all]
     $ cd examples
-    $ pip install -r requirements.txt
-    $ export FLASK_APP=app.py
-    $ flask db init
-    $ flask db create
-    $ flask index init
-    $ flask fixtures records
-    $ flask run --debugger
+    $ ./app-setup.sh
+    $ FLASK_APP=app.py flask fixtures records
+
+Run the server:
+
+.. code-block:: console
+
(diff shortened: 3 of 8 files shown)


hachreak:
ping @jirikuncar

jirikuncar:
@hachreak you can rebase now. Thanks!

hachreak:
ping @jirikuncar

examples/app-recreate.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
jirikuncar: What about calling the file `app-setup.sh` and having also `app-teardown.sh` with `flask db destroy --yes-i-know`?

examples/app-setup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+export FLASK_APP=app.py
+
+# Install specific dependencies
+pip install -r requirements.txt || exit 1
```
jirikuncar: What about using [`set -e`](http://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin) instead of `|| exit 1`?

hachreak:
@jirikuncar could you please have a look if it's ok.. them I'll change all PR for example apps as I did here

examples/app-setup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+set -e
```
jirikuncar: 👍

examples/app.py
```diff
@@ -27,18 +27,23 @@
...
+    $ pip install -e .[all]
     $ cd examples
-    $ pip install -r requirements.txt
-    $ export FLASK_APP=app.py
-    $ flask db init
-    $ flask db create
-    $ flask index init
-    $ flask fixtures records
-    $ flask run --debugger
+    $ # clean the environment if needed
```
jirikuncar: I would just move it at the end of the docstring so people don't loose the focus.

hachreak: As a ".. note::"?

jirikuncar: It's not really important since we are not including it in Sphinx.

examples/app-setup.sh
tiborsimko: If we standardise on shell scripts for testing example app, then it's good to add also an unbound symbol check, to be future-proof.  See e.g. [Invenio scripts](https://github.com/inveniosoftware/invenio/blob/master/scripts/populate-instance.sh#L98-L100) where I'm using:

```
# quit on errors:
set -o errexit

# quit on unbound symbols:
set -o nounset
```

But let me reply in the other shell-scripting-vs-Python-scripting thread...

.travis.yml
```diff
@@ -46,8 +47,6 @@ env:
 
 python:
   - "2.7"
-  - "3.3"
-  - "3.4"
```
tiborsimko: Let's remove 3.3 and 3.4 also from `CONTRIBUTING.rst`.

examples/app-setup.sh
jirikuncar: I was proposing this as an alternative to `|| exit 1` which was used before ...

.travis.yml
hachreak: done

PR #119, opened 2016-09-30
utils: new get_default_endpoint_for()

### LINKED ISSUE #101: RFC URL endpoint building

kaplun:
* NEW Introduces new RECORDS_REST_PID_TYPE_DEFAULT_ENDPOINT config
  variable, needed in order to know which default endpoint corresponds
  to which pid_type when building URIs. (closes #101)

Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

(supersedes #109)

### DIFF
--- a/invenio_records_rest/config.py
+++ b/invenio_records_rest/config.py
@@ -88,8 +88,9 @@ The structure of the dictionary is as follows:
 
 
     RECORDS_REST_ENDPOINTS = {
-        'record-pid-type': {
+        'endpoint-prefix': {
             'create_permission_factory_imp': permission_check_factory(),
+            'default_endpoint_prefix': True,
             'default_media_type': 'application/json',
             'delete_permission_factory_imp': permission_check_factory(),
             'item_route': ''/recods/<pid(record-pid-type):pid_value>'',
@@ -132,6 +133,10 @@ The structure of the dictionary is as follows:
 :param create_permission_factory_imp: Import path to factory that create
     permission object for a given record.
 
+:param default_endpoint_prefix: declare the current endpoint as the default
+    when building endpoints for the defined ``pid_type``. By default the
+    default prefix is defined to be the value of ``pid_type``.
+
 :param default_media_type: Default media type for both records and search.
 
 :param delete_permission_factory_imp: Import path to factory that creates a
@@ -146,10 +151,10 @@ The structure of the dictionary is as follows:
 :param max_result_window: Maximum total number of records retrieved from a
     query.
 
-:param pid_type: It specifies the record pid type. It's used also to build the
-    endpoint name. Required.
+:param pid_type: It specifies the record pid type. Required.
     You can generate an URL to list all records of the given ``pid_type`` by
-    calling ``url_for('invenio_records_rest.{0}_list'.format(pid_type))``.
+    calling ``url_for('invenio_records_rest.{0}_list'.format(
+    current_records_rest.default_endpoint_prefixes[pid_type]))``.
 
 :param pid_fetcher: It identifies the registered fetcher name. Required.
 
--- a/invenio_records_rest/utils.py
+++ b/invenio_records_rest/utils.py
@@ -41,42 +41,36 @@ from .errors import PIDDeletedRESTError, PIDDoesNotExistRESTError, \
 from .proxies import current_records_rest
 
 
-def build_default_endpoint_prefixes():
+def build_default_endpoint_prefixes(records_rest_endpoints):
     """Build the default_endpoint_prefixes map."""
-    ret = {}
-    record_rest_endpoints = current_app.config['RECORDS_REST_ENDPOINTS']
-    for endpoint in record_rest_endpoints.values():
+    pid_types = set()
+    guessed = set()
+    endpoint_prefixes = {}
+
+    for key, endpoint in records_rest_endpoints.items():
         pid_type = endpoint['pid_type']
-        ret[pid_type] = get_default_endpoint_for(pid_type,
-                                                 record_rest_endpoints)
+        pid_types.add(pid_type)
+        is_guessed = key == pid_type
+        is_default = endpoint.get('default_endpoint_prefix', False)
 
-    return ret
+        if is_default:
+            if pid_type in endpoint_prefixes and pid_type not in guessed:
+                raise ValueError('More than one "{0}" defined.'.format(
+                    pid_type
+                ))
+            endpoint_prefixes[pid_type] = key
+            guessed -= {pid_type}
+        elif is_guessed and pid_type not in endpoint_prefixes:
+            endpoint_prefixes[pid_type] = key
+            guessed |= {pid_type}
 
+    not_found = pid_types - set(endpoint_prefixes.keys())
+    if not_found:
+        raise ValueError('No endpoint-prefix for {0}.'.format(
+            ', '.join(not_found)
+        ))
 
-def get_default_endpoint_for(pid_type, _record_rest_endpoints=None):
-    """Get default endpoint for the given pid_type."""
-    if _record_rest_endpoints is None:
-        _record_rest_endpoints = current_app.config['RECORDS_REST_ENDPOINTS']
-
-    endpoint_prefix = None
-
-    for key, value in _record_rest_endpoints.items():
-        if (value['pid_type'] == pid_type and
-                value.get('default_endpoint_prefix')):
-            if endpoint_prefix is None:
-                endpoint_prefix = key
-            else:
-                raise ValueError('More than one endpoint-prefix has been '
-                                 'defined as default for '
-                                 'pid_type="{0}"'.format(pid_type))
-
-    if endpoint_prefix:
-        return endpoint_prefix
-    if pid_type in _record_rest_endpoints:
-        return pid_type
-
-    raise ValueError('No endpoint-prefix corresponds to pid_type="{0}"'.format(
-        pid_type))
+    return endpoint_prefixes
 
 
 def obj_or_import_string(value, default=None):
--- a/invenio_records_rest/utils.py
+++ b/invenio_records_rest/utils.py
@@ -38,6 +38,45 @@ from werkzeug.utils import cached_property, import_string
 from .errors import PIDDeletedRESTError, PIDDoesNotExistRESTError, \
     PIDMissingObjectRESTError, PIDRedirectedRESTError, \
     PIDUnregisteredRESTError
+from .proxies import current_records_rest
+
+
+def build_default_endpoint_prefixes():
+    """Build the default_endpoint_prefixes map."""
+    ret = {}
+    record_rest_endpoints = current_app.config['RECORDS_REST_ENDPOINTS']
+    for endpoint in record_rest_endpoints.values():
+        pid_type = endpoint['pid_type']
+        ret[pid_type] = get_default_endpoint_for(pid_type,
+                                                 record_rest_endpoints)
+
+    return ret
+
+
+def get_default_endpoint_for(pid_type, _record_rest_endpoints=None):
+    """Get default endpoint for the given pid_type."""
+    if _record_rest_endpoints is None:
+        _record_rest_endpoints = current_app.config['RECORDS_REST_ENDPOINTS']
+
+    endpoint_prefix = None
+
+    for key, value in _record_rest_endpoints.items():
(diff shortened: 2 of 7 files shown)


kaplun:
Travis is quite slow for inveniosoftware... I have submitted the PR an hour ago and has not yet started.

kaplun:
Just to note that this Travis failed on one of the builds due to unrelated ES timeout.

jacquerie:
> Just to note that this Travis failed on one of the builds due to unrelated ES timeout.

Are you sure? [This line](https://travis-ci.org/inveniosoftware/invenio-records-rest/jobs/163999297#L632) looks suspicious...

kaplun:
Ha! And only for Python 3.5?

kaplun:
Cc @jirikuncar, @lnielsen

invenio_records_rest/config.py
```diff
@@ -36,6 +36,16 @@
...
     return x
 
+RECORDS_REST_PID_TYPE_DEFAULT_ENDPOINT = dict(
+    recid='recid'
+)
+"""pid_type to default endpoint mapping.
+
+This mapping is needed in order to know which default endpoint corresponds to
+which pid_type when building URIs.
+"""
```
jacquerie:
```python
"""Mapping from a ``pid_type`` to its default endpoint.

This mapping is needed in order to know which endpoint name corresponds to
a ``pid_type`` when building URLs.
"""
```

kaplun:
Ready.

jacquerie:
LGTM, but my words carry no weight in these foreign lands.

kaplun:
Note this is currently a blocker for using upstream invenio-pidstore in INSPIRE. (due to the limitation of length of `pid_type`, and hence the need to decouple `endpoint`s and `pid_type`s, since in INSPIRE our endpoints are longer than 6 chars.

jirikuncar:
First, we should add more test for various use-cases - 2+ endpoints for single `pid_type`, no endpoint for `pid_type` (can it even happen?), ... (one `pid_type` can *not* point to multiple endpoints, can it?)

Then we should discuss again if we really need another configuration option and what it's default.

```python
def get_default_endpoint_for(pid_type):
    """Get default endpoint for the given pid_type."""
    return current_app.config.get(
        'RECORDS_REST_PID_TYPE_DEFAULT_ENDPOINT', {}
    ).get(pid_type, pid_type)
```

Would it be possible to mark an endpoint as default for given `pid_type`?

```python
RECORDS_REST_ENDPOINTS = dict(
    recid=dict(
        default_endpoint=True,
        pid_type='recid',
        ...
    )
)
```

Probably we don't have to mark the endpoint as default if endpoint name is equal to `pid_type`.

kaplun:
> First, we should add more test for various use-cases - 2+ endpoints for single pid_type, no endpoint for pid_type (can it even happen?), ... (one pid_type can not point to multiple endpoints, can it?)

Sure! What is the best pattern to test a function for different configuration values? I.e. how can I force a given configuration in a test?

> Then we should discuss again if we really need another configuration option and what it's default.

> ```python
def get_default_endpoint_for(pid_type):
    """Get default endpoint for the given pid_type."""
    return current_app.config.get(
        'RECORDS_REST_PID_TYPE_DEFAULT_ENDPOINT', {}
    ).get(pid_type, pid_type)
```
> Would it be possible to mark an endpoint as default for given pid_type?

> ```python
RECORDS_REST_ENDPOINTS = dict(
    recid=dict(
        default_endpoint=True,
        pid_type='recid',
        ...
    )
)
```
> Probably we don't have to mark the endpoint as default if endpoint name is equal to pid_type.

Yep! I also thought about just adding something like the `default_endpoint` flag. My only concerns with this are: 
* what if this flag is forgotten? We pick up first endpoint in case of ambiguity? We raise error/warning? 
* what if a distracted admin cut & paste twice the flag? We raise warning? Error?

jirikuncar:
> what if this flag is forgotten? We pick up first endpoint in case of ambiguity? We raise error/warning?

Probably we don't have to mark the endpoint as default if endpoint name is equal to pid_type.

> what if a distracted admin cut & paste twice the flag? We raise warning? Error?

Yes, the same we can do if there is no default key for `pid_type` (`pid_type not in RECORDS_REST_ENDPOINTS`) .

kaplun:
OK, there is just the pervers behavior of `endpoint` = `pid_type` but the default `endpoint` is set for another one. I guess in this case I will prioritize the default flag.

kaplun:
@jirikuncar what do you think about this new implementation?

invenio_records_rest/config.py
```diff
@@ -36,6 +36,7 @@
 def _(x):
     return x
 
+
```
kaplun: Should remove space.

invenio_records_rest/utils.py
```diff
@@ -139,7 +166,7 @@ def data(self):
                 response = make_response(jsonify(data), data['status'])
                 response.headers['Location'] = location
                 abort(response)
-            except BuildError:
+            except (BuildError, KeyError):
```
kaplun: Should amend for ValueError.

invenio_records_rest/utils.py
```diff
@@ -139,7 +166,7 @@ def data(self):
                 response = make_response(jsonify(data), data['status'])
                 response.headers['Location'] = location
                 abort(response)
-            except BuildError:
+            except (BuildError, ValueError):
```
jirikuncar: Maybe we should make sure that the configuration is check when the extension is initialized instead of catching it runtime.

kaplun: There is still though the possibility that some client code specify a `pid_type` that is not defined. In this case `ValueError` would still be triggered.

invenio_records_rest/utils.py
```diff
@@ -40,6 +40,33 @@
     PIDUnregisteredRESTError
 
 
+@cached_property
```
kaplun: OK. Just for understanding better: why it would be wrong to put it here?

invenio_records_rest/utils.py
jirikuncar: Do you have an example? Probably it should be check already in `build_pid_type_redirects` ...

invenio_records_rest/utils.py
kaplun: Oh. What you propose would entail a full-reimplementation though.

jirikuncar: The question is if we then need `get_default_endpoint_for` function instead using the property e.g. `endpoint_for_pid_type` directly so the code will look: `current_records_rest.endpoint_for_pid_type[pid_type]`

(other possible namings: `default_endpoints`, `pid_type_redirects`, `endpoints_for`, ...)

kaplun: What about `default_endpoint_prefixes`? (since it's not actually a endpoint what is provided but just the prefix to be used to build the set of end points, `{0}_list`, `{0}_item`...)

invenio_records_rest/utils.py
kaplun: OK, In the last implementation the whole mapping is built at module initialization (or anyway first call of the `cached_property`, so all inconsistencies should be good.

kaplun:
@jirikuncar again ready

invenio_records_rest/utils.py
```diff
@@ -38,6 +38,45 @@
...
+    record_rest_endpoints = current_app.config['RECORDS_REST_ENDPOINTS']
+    for endpoint in record_rest_endpoints.values():
+        pid_type = endpoint['pid_type']
+        ret[pid_type] = get_default_endpoint_for(pid_type,
+                                                 record_rest_endpoints)
+
+    return ret
+
+
+def get_default_endpoint_for(pid_type, _record_rest_endpoints=None):
```
jirikuncar: do you still need it?

tests/test_utils.py
```diff
@@ -0,0 +1,104 @@
...
+from __future__ import absolute_import, print_function
+
+from pytest import raises
+
+from invenio_records_rest.utils import get_default_endpoint_for
+
+
+def test_get_default_endpoint_for():
+    """Test get_default_endpoint_for()."""
+    assert get_default_endpoint_for('recid', {
```
jirikuncar: The tests should be probably adapted to the refactored code.

jirikuncar: can we check `current_records_rest.default_endpoint_prefixes` instead?

kaplun: Testing `get_default_endpoint_for` was my solution in order to play with different configurations (that I pass as argument), while keeping the code simple.

Otherwise I would have to mock different `RECORDS_REST_ENDPOINTS`, but I don't know how to do it. 

Overall the coverage should be identical.

jirikuncar: @kaplun you can make the test `@pytest.mark.parametrized`

jirikuncar: > Overall the coverage should be identical.

Yes, but here we are testing "low-level" function which might prevent someone from optimizing the build process for `default_endpoint_prefixes`.

tests/conftest.py
```diff
@@ -156,6 +171,9 @@ def test_mytest(app, db, es):
         if 'endpoint' in request.param:
             app.config['RECORDS_REST_ENDPOINTS']['recid'].update(
                 request.param['endpoint'])
+        elif 'records_rest_endpoints' in request.param:
```
jirikuncar: one can combine both, hence just `if` instead of `elif`

tests/conftest.py
```diff
@@ -112,6 +112,21 @@ def test_mytest(app, db, es):
...
     RECORDS_REST_ENDPOINTS.
+
+    Alternatively:
+
+    .. code-block:: python
+
+    @pytest.mark.parametrize('app', [dict(
+        records_rest_endpoints=dict(
+            recid=dict(
+                search_class='conftest:TestSearch',
```
jirikuncar: copy&paste issue?

tests/conftest.py
kaplun: Kind of, because `records_rest_endpoints` somehow overrides the weaker `endpoint`. But OK.

kaplun:
@jirikuncar your patch LGTM

kaplun:
Yep build still failing due to what I was mentioning you on gitter: i.e. that the cached_property is leaking through the tests. (in fact if you locally launch `py.tests tests/test_views_options.py` it will complete successfully.

jirikuncar:
LGTM now :)

tests/conftest.py
```diff
@@ -112,13 +113,28 @@ def test_mytest(app, db, es):
...
+
+    This will fully parameterize RECORDS_REST_ENDPOINTS.
     """
     instance_path = tempfile.mkdtemp()
     app = Flask('testapp', instance_path=instance_path)
     app.config.update(
         INDEXER_DEFAULT_DOC_TYPE='testrecord',
         INDEXER_DEFAULT_INDEX=search_class.Meta.index,
-        RECORDS_REST_ENDPOINTS=config.RECORDS_REST_ENDPOINTS,
+        RECORDS_REST_ENDPOINTS=copy.deepcopy(config.RECORDS_REST_ENDPOINTS),
```
kaplun: Ha! Well done @jirikuncar

PR #123, opened 2016-10-07
search: error handling

### LINKED ISSUE #122: search: error handling
Elasticsearch Queries that include [reserved characters](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-query-string-query.html#_reserved_characters) are rejected and lead to a 400 response from ES. However in some Elasticsearch installations/versions the error type that Elasticsearch is including in the response is not the same as the one handled by [the error handler](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L68).

On Elasticsearch v2.3.5:
`http :9200/records/_search?q=i%2Fo` >> `error.root_cause.type: "query_parsing_exception"`

On Elasticsearch v2.3.0:
`http :9200/records/_search?q=i%2Fo` >> `error.root_cause.type: "token_mgr_error"`

Possible solutions (besides updating to latest Elasticsearch version) are either extending the dictionary with more error codes, or adding it to the app configuration.

slint:
* BETTER Moves Elasticsearch error handling to application config,
  allowing for wider coverage of error codes. (closes #122)

Signed-off-by: Alexander Ioannidis <a.ioannidis@cern.ch>

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -64,15 +64,13 @@ def elasticsearch_query_parsing_exception_handler(error):
     return InvalidQueryRESTError(description=description).get_response()
 
 
-elasticsearch_error_handlers = {
-    'query_parsing_exception': elasticsearch_query_parsing_exception_handler,
-}
-"""List of handlers for ElasticSearch errors."""
-
-
 def create_error_handlers(blueprint):
-    """Create error handlers on blueprint."""
-    # catch record validation errors
+    """Create error handlers on blueprint.
+
+    :params blueprint: Records API blueprint.
+    :returns: Configured blueprint.
+    """
+    # Catch record validation errors
     @blueprint.errorhandler(ValidationError)
     def validation_error(error):
         """Catch validation errors."""
@@ -81,17 +79,14 @@ def create_error_handlers(blueprint):
     @blueprint.errorhandler(RequestError)
     def elasticsearch_badrequest_error(error):
         """Catch errors of ElasticSearch."""
-        def first(function, iterable, default=None):
-            """Return the first item from iterable which function is true."""
-            for item in iterable:
-                if function(item):
-                    return item
-            return default
+        handlers = current_app.config[
+            'RECORDS_REST_ELASTICSEARCH_ERROR_HANDLERS']
+        cause_types = {c['type'] for c in error.info['error']['root_cause']}
 
-        handlers = map(lambda c: elasticsearch_error_handlers.get(c['type']),
-                       error.info['error']['root_cause'])
-        handler = first(lambda h: h, handlers, lambda h: h)
-        return handler(error)
+        for cause_type, handler in handlers.items():
+            if cause_type in cause_types:
+                return handler(error)
+        return error
 
     return blueprint
 
@@ -99,7 +94,7 @@ def create_error_handlers(blueprint):
 def create_blueprint(endpoints):
     """Create Invenio-Records-REST blueprint.
 
-    :params: Dictionary representing the endpoints configuration.
+    :params endpoints: Dictionary representing the endpoints configuration.
     :returns: Configured blueprint.
     """
     blueprint = Blueprint(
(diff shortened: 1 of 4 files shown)


PXke:
Seems to be linked to https://github.com/inveniosoftware/invenio-records-rest/issues/122

slint:
ping @jirikuncar

tests/test_views_search.py
```diff
@@ -226,3 +230,16 @@ def test_query_wrong(app, indexed_records, search_url):
...
         res = client.get(search_url, query_string={'q': 'test;'})
         assert res.status_code == 200
+
+
+def test_elasticsearch_exception(app, indexed_records, search_url):
+    with app.test_client() as client:
+        for error_type in elasticsearch_error_handlers.keys():
+            es_req_error = RequestError(
+                '', '', {'error': {'root_cause': [{'type': error_type}]}})
+            with patch('invenio_records_rest.views.RecordsListResource.get',
```
jirikuncar: Can you make a request that really throws the exception instead of just patching the get method? Otherwise we won't be able to detect that the exception name has changed.

slint:
@jirikuncar I added the Elasticsearch exceptions in the app configuration, although I will also have to update `invenio-deposit`, which uses [`invenio_records_rest.views.create_error_handlers`](https://github.com/slint/invenio-records-rest/blob/error_handling/invenio_records_rest/views.py#L67-L91) ([here](https://github.com/inveniosoftware/invenio-deposit/blob/master/invenio_deposit/views/rest.py#L62)), and would fail on app initialization.

invenio_records_rest/views.py
```diff
@@ -64,42 +64,39 @@ def elasticsearch_query_parsing_exception_handler(error):
     return InvalidQueryRESTError(description=description).get_response()
 
 
-elasticsearch_error_handlers = {
```
jirikuncar: You can change this variable to be a proxy that resolves the `RECORDS_REST_ELASTICSEARCH_ERROR_HANDLERS ` configuration option without changing the API of `create_error_handlers`.

slint: My plan was actually to do the `obj_or_import_string` call during app initialization, so that if for some reason the config is wrong or the import fails, the application won't start at all, instead of failing when the time comes for the specific error code to be handled.

slint: @jirikuncar I think I found a middle-ground which keeps compatibility with `invenio-deposit`: doing the `obj_or_import_string` inside the `init_config`.

invenio_records_rest/views.py
```diff
@@ -81,25 +79,20 @@ def validation_error(error):
...
 
-    return blueprint
-
 
 def create_blueprint(endpoints):
     """Create Invenio-Records-REST blueprint.
 
-    :params: Dictionary representing the endpoints configuration.
+    :params endpoints: Dictionary representing the endpoints configuration.
+    :params es_error_handlers: Dictionary with Elasticsearch error handlers for
```
jirikuncar: please remove this comment

invenio_records_rest/views.py
```diff
@@ -111,8 +104,8 @@ def create_blueprint(endpoints):
     for endpoint, options in (endpoints or {}).items():
         for rule in create_url_rules(endpoint, **options):
             blueprint.add_url_rule(**rule)
-
-    return create_error_handlers(blueprint)
```
jirikuncar: please keep it

invenio_records_rest/views.py
```diff
@@ -81,25 +79,20 @@ def validation_error(error):
...
-                    return item
-            return default
-
-        handlers = map(lambda c: elasticsearch_error_handlers.get(c['type']),
-                       error.info['error']['root_cause'])
-        handler = first(lambda h: h, handlers, lambda h: h)
+        handlers = current_app.config[
+            'RECORDS_REST_ELASTICSEARCH_ERROR_HANDLERS']
+        cause_types = {c['type'] for c in error.info['error']['root_cause']}
+        handler = next((v for k, v in handlers.items()
```
jirikuncar: I find it quite dificult to read. WDYT about:
```python
for cause_type, handler in handlers.items():
    if cause_type in cause_types:
        return handler(error)
return error
```

slint:
@jirikuncar Any other changes required for this PR?

jirikuncar:
LGTM

PR #134, opened 2016-11-21
global: Datacite v4.0 support.

remileduc:
Added wrapper class for the new Datacite v4.0 support in Datacite module (see https://github.com/inveniosoftware/datacite/pull/18)

INCOMPATIBLE: the constructor of `OAIDataCiteSerializer` now takes a "serializer" parameter instead of a "v31".

Signed-off-by: Rémi Ducceschi <remi.ducceschi@gmail.com>

### DIFF
--- a/invenio_records_rest/serializers/datacite.py
+++ b/invenio_records_rest/serializers/datacite.py
@@ -26,7 +26,7 @@
 
 from __future__ import absolute_import, print_function
 
-from datacite import schema31
+from datacite import schema31, schema40
 from invenio_records.api import Record
 from lxml import etree
 from lxml.builder import E
@@ -34,12 +34,17 @@ from lxml.builder import E
 from .marshmallow import MarshmallowSerializer
 
 
-class DataCite31Serializer(MarshmallowSerializer):
+class BaseDataCiteSerializer(MarshmallowSerializer):
     """Marshmallow based DataCite serializer for records.
 
     Note: This serializer is not suitable for serializing large number of
     records.
     """
+
+    schema = None
+    """Class variable to define which schema use."""
+    version = None
+    """Class variable to tell the version of the schema."""
 
     def serialize(self, pid, record, links_factory=None):
         """Serialize a single record and persistent identifier.
@@ -48,7 +53,7 @@ class DataCite31Serializer(MarshmallowSerializer):
         :param record: Record instance.
         :param links_factory: Factory function for record links.
         """
-        return schema31.tostring(
+        return self.schema.tostring(
             self.transform_record(pid, record, links_factory))
 
     def serialize_search(self, pid_fetcher, search_result, links=None,
@@ -61,7 +66,7 @@ class DataCite31Serializer(MarshmallowSerializer):
         """
         records = []
         for hit in search_result['hits']['hits']:
-            records.append(schema31.tostring(self.transform_search_hit(
+            records.append(self.schema.tostring(self.transform_search_hit(
                 pid_fetcher(hit['_id'], hit['_source']),
                 hit,
                 links_factory=item_links_factory,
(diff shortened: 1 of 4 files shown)


invenio_records_rest/serializers/datacite.py
```diff
@@ -75,7 +79,31 @@ def serialize_oaipmh(self, pid, record):
...
+    Note: This serializer is not suitable for serializing large number of
+    records.
+    """
+
+    """Class variable to define which schema use."""
+    schema = schema40
+    version = '4.0'
 
 
 class OAIDataCiteSerializer(object):
```
remileduc: Here, I left the parameter "v31" in the `__init__` method for backward compatibility. But it may be better to change it as it may no longer be the v31 of Datacite...

lnielsen: We can change it to:

```python
class OAIDataCiteSerializer(object):
    def __init__(self, serializer=None, datacentre=None, is_reference_quality='true'):
        self.serializer = serializer
        self.datacentre = datacentre
        self.is_reference_quality = is_reference_quality
```

Just add an ``INCOMPATIBLE`` to the commit message.

tests/test_serializer_datacite.py
```diff
@@ -93,13 +94,25 @@ def fetcher(obj_uuid, data):
...
 
-    tree = OAIDataCiteSerializer(v31=s, datacentre='CERN').serialize_oaipmh(
+    tree = OAIDataCiteSerializer(serializer=s,
+                                 datacentre='CERN').serialize_oaipmh(
         PersistentIdentifier(pid_type='doi', pid_value='10.1234/b'),
         {'_source': dict(doi='10.1234/b'), '_id': 'b', '_version': 1})
     assert len(tree.xpath('/oai_datacite/datacentreSymbol')) == 1
+
+
+def test_serialize():
```
jirikuncar: You can solve it by making `test_serialize_search` parametrized.

```python
@pytest.mark.parametrized('serializer', [(DataCite31Serializer, ), (DataCite40Serializer, )])
def test_serialize_search(serializer):
    ...
    data = serializer(SimpleSchema).serialize_search(
```

remileduc: nice feature! I've updated the code.

invenio_records_rest/serializers/datacite.py
```diff
@@ -26,29 +26,34 @@
...
+class BaseDataCiteSerializer(MarshmallowSerializer):
     """Marshmallow based DataCite serializer for records.
 
     Note: This serializer is not suitable for serializing large number of
     records.
     """
 
+    schema = None
+    """Class variable to define which schema use."""
+    version = 'Unknown'
```
jirikuncar: IMHO it would be more understandable to use `None` (or empty string). Basically something that evaluates to `False`.

invenio_records_rest/serializers/datacite.py
```diff
@@ -75,15 +80,42 @@ def serialize_oaipmh(self, pid, record):
...
+    """Class variable to define which schema use."""
+    version = '4.0'
+    """Class variable to tell the version of the schema."""
 
 
 class OAIDataCiteSerializer(object):
     """OAI DataCite serializer (only for OAI-PMH)."""
 
-    def __init__(self, v31=None, datacentre=None, is_reference_quality='true'):
+    def __init__(self, serializer=None, datacentre=None,
```
lnielsen: Each instance has to provide their own DataCite serializer (which will be either v31 or v4). See e.g. current [usage example](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/__init__.py#L77)

lnielsen:
LGTM

PR #139, opened 2017-03-13
errors: adds jsonschema validation class with error message

ioannistsanaktsidis:
Signed-off-by: Ioannis Tsanaktsidis <ioannis.tsanaktsidis@cern.ch>

### DIFF
--- a/invenio_records_rest/errors.py
+++ b/invenio_records_rest/errors.py
@@ -27,7 +27,7 @@
 from __future__ import absolute_import, print_function
 
 from flask import request
-from invenio_rest.errors import RESTException
+from invenio_rest.errors import RESTException, RESTValidationError
 
 
 #
@@ -177,3 +177,15 @@ class SuggestNoCompletionsRESTError(RESTException):
         super(RESTException, self).__init__(**kwargs)
         self.description = 'No completions requested.{0}'.format(
             ' (options: {0})'.format(options) if options else '')
+
+
+class JSONSchemaValidationError(RESTValidationError):
+    """JSONSchema validation error exception."""
+
+    code = 400
+
+    def __init__(self, error=None, **kwargs):
+        """Error description."""
+        super(RESTValidationError, self).__init__(**kwargs)
+        self.description = 'Validation error: {0}.'.format(
+            error.message if error else '')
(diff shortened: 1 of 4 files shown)


lnielsen:
Looks good, but can you add a small test. 

I think you can add the test in [test_views_list_post.py](https://github.com/inveniosoftware/invenio-records-rest/blob/master/tests/test_views_list_post.py). To trigger a schema validation error I think you can simply dump a simple schema directly in ``$schema`` instead of having to reference a URL. 

```
record = {
    'title': 1,
    '$schema': {
        'properties': {
            'title': {'type': 'string'}
        }
    }
}

File: tests/test_views_list_post.py [hunk 1]
```diff
@@ -140,3 +140,29 @@ def test_validation_error(app, db, test_data, search_url, content_type):
         res = client.post(
             search_url, data=json.dumps(test_data[0]), headers=HEADERS)
         assert res.status_code == 400
+
+
+@mock.patch('invenio_records.api.Record.validate', _mock_validate_fail)
```
lnielsen: If this is here, doesn't it mean that the JSON schema is actually not being validated? I mean if you remove, I think it should still fail, and it would test better what you need.

tests/test_views_list_post.py
```diff
@@ -140,3 +140,29 @@ def test_validation_error(app, db, test_data, search_url, content_type):
...
+    with app.test_client() as client:
+        HEADERS = [
+            ('Accept', 'application/json'),
+            ('Content-Type', content_type)
+        ]
+
+        # Create record
+        res = client.post(
+            search_url, data=json.dumps(record), headers=HEADERS)
+        assert res.status_code == 400
```
lnielsen: Specifically in this case, I would also test the content of the response, since this line will pass with the old exception class as well. I.e. something a long the lines of :

```assert res.json['message']```

File: invenio_records_rest/errors.py [hunk 3]
```diff
@@ -177,3 +177,14 @@ def __init__(self, options=None, **kwargs):
...
         self.description = 'No completions requested.{0}'.format(
             ' (options: {0})'.format(options) if options else '')
+
+
+class JSONSchemaValidationError(RESTValidationError):
+    """JSONSchema validation error exception."""
+
+    code = 400
+
+    def __init__(self, description=None, error=None, **kwargs):
```
egabancho: Do you really need the description here? It feels a bit odd 😉 

Also something like this might be more reliable

```
self.description = 'Validation error: {0}.'.format(error.message if error else '')
```

lnielsen:
Regarding build errors see inveniosoftware/troubleshooting#9

PR #142, opened 2017-03-27
serialization: JSON-LD serializer

jma:
Signed-off-by: Johnny Mariéthoz <Johnny.Mariethoz@rero.ch>

### DIFF
--- a/invenio_records_rest/serializers/jsonld.py
+++ b/invenio_records_rest/serializers/jsonld.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2017 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


lnielsen:
@jma I've made some few updates in the code. Could you please have a look.

I understand that by default you just want to get ``obj['metadata']``, then attach the context and last run it through JSON-LD expansion. 

I'm not really sure how to handle nicely. Right now, I've added a marshmallow schema that by default does that - what I don't like about this, is the overhead of marshmallow just to grab an attribute. OTOH, if ``dump()`` only grabs the ``metadata`` then the JSON-LD serializer will never be able to used ``created``, ``updated`` etc

jma:
Thanks Lars,

is this `json-ld` the first serialiser that use only metadata? `expanded` and `compacted` is ok for me.

Else all is fine for me, I just made a small change to avoid tests warning.

lnielsen:
@jma I think it's the only serializer so far. The problem by using only metadata is that you don't have access to the created and updated date as well as the PID by which the record was accessed.

PR #148, opened 2017-05-23
links: additional links

### LINKED ISSUE #46: views: html link in json serialization
**Problem**:
We show search results with a angular application (this is the [template used](https://github.com/inveniosoftware/invenio-search-ui/blob/master/invenio_search_ui%2Fstatic%2Ftemplates%2Finvenio_search_ui%2Fmarc21%2Fdefault.html#L3)): it receives the query results from `invenio-records-rest` json serializer that it's a list of records from elasticsearch, "improved" with some extra fields inserted by the serializer.
Currently, inside the json we have a field [links](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest%2Flinks.py#L33) that contains the url that point to the json version of the record (something like `/api/records/[:recid]`).
But, what we need is to have also a field `html_url` that contains the link to the "html version" of the record.

To be able to change the template as:

``` javascript
<h4>
  <a target="_self"
     ng-href="{{ record.links.html_url }}">
    {{ record.metadata.title_statement.title }}
   </a>
</h4>
```

---

cc @lnielsen @drjova @tiborsimko @jirikuncar @krzysztof

nikofil:
* Adds the capability to specify additional links for a record. (closes #46)

### DIFF
--- a/invenio_records_rest/links.py
+++ b/invenio_records_rest/links.py
@@ -24,7 +24,7 @@
 
 """Links for record serialization."""
 
-from flask import url_for
+from flask import request, url_for
 
 from .proxies import current_records_rest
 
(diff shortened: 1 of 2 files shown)


drjova:
@nikofil could you please rebase?

Issue #131, opened 2016-11-03
views: get the search_index from the config variable


zazasa:
The `search_index` option set in the `RECORDS_REST_ENDPOINTS` is not respected by the search api.
Indeed the search_index seems to be always equal to "_all" .
I verified that hardcoding the index here https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L436 :
```python
search = self.search_class(index='test-index').params(version=True)
```
seems working, but i had some problem reading this parameter from the config.

jirikuncar:
@zazasa @audub are you planning to submit a PR?

zazasa:
I am not sry, it is not clear enough for me to fix it. Maybe we can talk about it in these days so I can do it.

dinosk:
@zazasa I could not reproduce this, I noted though that the search index registered will be the one that is there at the [initialization](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/ext.py#L107) of the module. Changing it afterwards won't work.

PR #150, opened 2017-08-08
views: specify search index

dinosk:
(closes https://github.com/inveniosoftware/invenio-records-rest/issues/131)

Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -433,7 +433,13 @@ class RecordsListResource(ContentNegotiatedMethodView):
 
         # Arguments that must be added in prev/next links
         urlkwargs = dict()
-        search = self.search_class().params(version=True)
+        self.search_index = current_app.config[
+            'RECORDS_REST_ENDPOINTS'][self.pid_type]['search_index']
+        if self.search_index:
+            search = self.search_class(index=self.search_index).\
+                params(version=True)
+        else:
+            search = self.search_class().params(version=True)
         search = search[(page - 1) * size:page * size]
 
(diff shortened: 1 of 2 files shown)


invenio_records_rest/views.py
```diff
@@ -433,7 +433,13 @@ def get(self, **kwargs):
 
         # Arguments that must be added in prev/next links
         urlkwargs = dict()
-        search = self.search_class().params(version=True)
+        self.search_index = current_app.config[
+            'RECORDS_REST_ENDPOINTS']['recid']['search_index']
```
lnielsen: ``recid`` -> ``self.pid_type``

invenio_records_rest/views.py
```diff
@@ -433,7 +433,13 @@ def get(self, **kwargs):
 
         # Arguments that must be added in prev/next links
         urlkwargs = dict()
-        search = self.search_class().params(version=True)
+        self.search_index = current_app.config[
```
lnielsen: I not sure I understand why ``search_class`` should not include the index. Above a partial is created which should have the keyword arguments: https://github.com/dinosk/invenio-records-rest/blob/53023c55077dc43b1b643e4b460b6c529d3ffdd7/invenio_records_rest/views.py#L212

dinosk: This makes more sense now, the only problem I can think of is that someone tries to modify the search_index in the config.py after partial has already been called and the changes are not registered. If this is not supported then these changes aren't needed and I can close this.

PR #153, opened 2017-09-21
facets: dynamics aggregations addition

hachreak:


hachreak:
With this PR merged we should be able to use original aggregation and default facets factory.
It means remove this [aggregation](https://github.com/CERNDocumentServer/cds-videos/blob/cdslabs_qa/cds/modules/deposit/facets.py#L42-L51) code and personalized [facets factory](https://github.com/CERNDocumentServer/cds-videos/blob/cdslabs_qa/cds/modules/deposit/facets.py#L64-L82).

------------------------------------------------------
cc @egabancho @drjova

PR #154, opened 2017-10-03
global: param for serializer aliases query arg

ntarocco:


invenio_records_rest/views.py
```diff
@@ -117,6 +117,7 @@ def create_url_rules(endpoint, list_route=None, item_route=None,
                      delete_permission_factory_imp=None,
                      record_class=None,
                      record_serializers=None,
+                     record_serializers_query_aliases=None,
```
slint: Maybe just `(record_)serializers_aliases`, to keep it generic? (in case somebody wants to something fancy...)

ntarocco: done!

ntarocco:
@nharraud see https://github.com/inveniosoftware/invenio-rest/pull/78 for to have a better understanding of the change.
The aim is to be able to call
`/api/record/<id>?format=json`
to be able to export correctly a record using a redirect without headers

invenio_records_rest/views.py
```diff
@@ -117,6 +117,7 @@ def create_url_rules(endpoint, list_route=None, item_route=None,
                      delete_permission_factory_imp=None,
```
nharraud: Can you fix the copyright year in this file please?

ntarocco: Absolutely, done. Sorry I didn't notice it here.
I have also added the release commit.

PR #166, opened 2017-11-23
global: refactor serializers

### LINKED ISSUE #110: Make the use of marhsmallow optionnal in serializers
Sometimes using `marhsmallow` in serializers can be overkill. We can make the schema optional in `invenio_record_rest.serializers.json:JSONSerializer`. Similar to [https://github.com/inveniosoftware/invenio-marc21/blob/master/invenio_marc21/serializers/marcxml.py#L54](url).

nharraud:
* NEW Adds support for marshmallow context in MarshmallowMixin.
  (closes #110)

* BETTER Refactors the serializers in order to make them work
  better as mixins.

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/invenio_records_rest/links.py
+++ b/invenio_records_rest/links.py
@@ -33,7 +33,7 @@ from flask import request, url_for
 from .proxies import current_records_rest
 
 
-def default_links_factory(pid):
+def default_links_factory(pid, record=None, **kwargs):
     """Factory for record links generation.
 
     :param pid: A Persistent Identifier instance.
@@ -53,7 +53,7 @@ def default_links_factory_with_additional(additional_links):
            returned object.
     :returns: A link generation factory.
     """
-    def factory(pid):
+    def factory(pid, **kwargs):
         links = default_links_factory(pid)
         for link in additional_links:
             links[link] = additional_links[link].format(pid=pid,
--- a/invenio_records_rest/serializers/base.py
+++ b/invenio_records_rest/serializers/base.py
@@ -31,22 +31,137 @@ import copy
 import pytz
 
 
-class PreprocessorMixin(object):
+class SerializerMixinInterface(object):
+    """Mixin serializing records.
+
+    A record serializer should inherit subclasses of
+    - SerializerMixinInterface
+    - TransformerMixinInterface
+    - PreprocessorMixinInterface
+
+    This class forwards its input to the transformer mixin and serializes its
+    result.
+    """
+
+    def serialize(self, pid, record, links_factory=None, **kwargs):
+        """Serialize a single record and persistent identifier.
+
+        This method should delegate the record transformation to the
+        transformer mixin's transform_record method.
+
+        :param pid: Persistent identifier instance.
+        :param record: Record instance.
+        :param links_factory: Factory function for record links.
+        """
+        raise NotImplementedError()
+
+    def serialize_search(self, pid_fetcher, search_result, links=None,
+                         item_links_factory=None, **kwargs):
+        """Serialize a search result.
+
+        This method should delegate each record search hit transformation
+        to the transformer mixin's transform_search_hit method.
+
+        :param pid_fetcher: Persistent identifier fetcher.
+        :param search_result: Elasticsearch search result.
+        :param links: Dictionary of links to add to response.
+        :param item_links_factory: Factory function for record links.
+        """
+        raise NotImplementedError()
+
+    def serialize_oaipmh(self, pid, record):
+        """Serialize a single record for OAI-PMH.
+
+        :param pid: The :class:`invenio_pidstore.models.PersistentIdentifier`
+            instance.
+        :param record: The :class:`invenio_records.api.Record` instance.
+        :returns: The object serialized.
+        """
+        raise NotImplementedError()
+
+
+class TransformerMixinInterface(object):
+    """Mixin transforming records during serialization.
+
+    This class should be mixed with a PreprocessorMixinInterface subclass and
+    a SerializerMixinInterface subclass.
+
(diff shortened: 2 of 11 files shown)


nharraud:
@lnielsen this PR refactors the MarshmallowSerializer as a Mixin. If somebody wants to have a serializer without schemas they can implement their own transformation mixin.
I can also transform all the other serializers as mixin.

The serializers still exist, they subclass all the necessary mixins. I transformed the JSON serializer as a mixin too. I can do the same for the other serializers if you want.

The classes should be mostly backward compatible.

Note that I also modified the link_factory use. It now receives all the parameters from the serializer. This avoids a problem we face in B2SHARE were we need the record metadata in order to generate some links. I think there is the same problem in Zenodo.

invenio_records_rest/links.py
```diff
@@ -29,7 +29,7 @@
 from .proxies import current_records_rest
 
 
-def default_links_factory(pid):
+def default_links_factory(pid, **kwargs):
```
slint: As you mentioned, this seems to be the only breaking change, considering that because of the previous "non-kwargs" signature, overlay link factories where defined having in mind the single argument definition...

Would it be possible to have this work for old functions (eg. [through inspection](https://stackoverflow.com/a/33029621)) but trigger an annoying warning?

*(PS: We will, of course, update our link factories on Zenodo, but I'm not sure how other people handle package upgrades)*

nharraud:
@slint I modified the create_blueprint function in order to wrap any old-signature links_factory. Can you take a look?

invenio_records_rest/serializers/base.py
```diff
@@ -34,19 +34,20 @@
...
 
-    def __init__(self, replace_refs=False):
-        """."""
+    def __init__(self, replace_refs=False, **kwargs):
+        """Constructor."""
+        super(PreprocessorMixin, self).__init__(**kwargs)
         self.replace_refs = replace_refs
 
-    def preprocess_record(self, pid, record, links_factory=None):
+    def preprocess_record(self, pid, record, links_factory=None, **kwargs):
```
lnielsen: What are the ``kwargs`` needed for?

invenio_records_rest/serializers/marshmallow.py
```diff
@@ -29,25 +29,28 @@
 from .base import PreprocessorMixin
 
 
-class MarshmallowSerializer(PreprocessorMixin):
```
lnielsen: I would keep ``MarshmallowSerializer`` to avoid nasty surprises that it has been removed (i.e. like you have ``JSONLDSerializer``).

invenio_records_rest/serializers/base.py
```diff
@@ -31,22 +31,137 @@
...
+
+    def serialize_oaipmh(self, pid, record):
+        """Serialize a single record for OAI-PMH.
+
+        :param pid: The :class:`invenio_pidstore.models.PersistentIdentifier`
+            instance.
+        :param record: The :class:`invenio_records.api.Record` instance.
+        :returns: The object serialized.
+        """
+        raise NotImplementedError()
```
nharraud: @lnielsen Here is the `serialize_oaipmh` method. It is not implemented by the JSONSerializer as it does not support OAIPMH

invenio_records_rest/serializers/marshmallow.py
```diff
@@ -26,28 +26,35 @@
...
+from .base import TransformerMixinInterface
+from .schemas.json import RecordSchemaJSONV1
 
 
-class MarshmallowSerializer(PreprocessorMixin):
+class MarshmallowMixin(TransformerMixinInterface):
     """Base class for marshmallow serializers."""
 
-    def __init__(self, schema_class=None, replace_refs=False):
+    def __init__(self, schema_class=RecordSchemaJSONV1, **kwargs):
```
nharraud: @lnielsen I set the default schema as RecordSchemaJSONV1. It is equivalent to not giving a schema with `PreprocessorMixin`. Only `revision` is filtered out.

invenio_records_rest/serializers/base.py
lnielsen: I think that's fine. If used it will correctly say that it's not compatible.

invenio_records_rest/serializers/jsonld.py
```diff
@@ -29,31 +29,19 @@
...
-    """Schema which extracts the metadata only."""
-
-    @post_dump
-    def post_dump(self, obj):
-        """Get metadata only from the record."""
-        return obj.get('metadata', {})
-
-
-class JSONLDSerializer(JSONSerializer):
+class JSONLDSerializerMixin(object):
```
lnielsen: Should this inherit from ``TransformerMixinInterface``

invenio_records_rest/serializers/dc.py
```diff
@@ -29,10 +29,11 @@
 from dcxml import simpledc
 from invenio_records.api import Record
 
-from .marshmallow import MarshmallowSerializer
+from .base import PreprocessorMixin
+from .marshmallow import MarshmallowMixin
 
 
-class DublinCoreSerializer(MarshmallowSerializer):
+class DublinCoreSerializer(MarshmallowMixin, PreprocessorMixin):
```
lnielsen: Should this inherit from ``SeriazlierMixinInterface``?

invenio_records_rest/views.py
```diff
@@ -58,6 +58,23 @@
 from .query import es_search_factory
 from .utils import obj_or_import_string
 
+try:  # Python 3 way of inspecting functions
```
lnielsen: Can we move this to a ``_compat.py`` module?

PR #167, opened 2017-11-23
docs: restructuring API docs

krzysztof:
Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -28,23 +28,32 @@ API Docs
 .. automodule:: invenio_records_rest.ext
    :members:
 
-Configuration
--------------
-
-.. automodule:: invenio_records_rest.config
-   :members:
+Facets
+------
 
 .. automodule:: invenio_records_rest.facets
    :members:
 
+Sorter
+------
+
 .. automodule:: invenio_records_rest.sorter
    :members:
+
+Links
+-----
 
 .. automodule:: invenio_records_rest.links
    :members:
 
+Query
+-----
+
 .. automodule:: invenio_records_rest.query
    :members:
+
+Utils
+-----
 
 .. automodule:: invenio_records_rest.utils
    :members:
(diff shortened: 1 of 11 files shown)


docs/api.rst
```diff
@@ -59,5 +68,24 @@ Views
...
+.. autofunction:: invenio_records_rest.views.need_record_permission
+
+each resource method is checking the permissions through this decorator, with
+``read_permission_factory``, ``create_permission_factory``,
+``update_permission_factory`` and ``delete_permission_factory`` as argument,
+where the implementation of each can be configured for each REST
+endpoint with :data:`invenio_records_rest.config.RECORDS_REST_ENDPOINTS`
+(using ``{read,create,update,delete}_permission_factory_imp`` keys).
+
+For reference on each resource method and its the default permission factory,
```
slint: *and its ~the~ default permission*

invenio_records_rest/utils.py
```diff
@@ -187,10 +192,18 @@ def data(self):
...
 
 class PIDConverter(BaseConverter):
-    """Resolve PID value."""
+    """Converter for PID values in the route mapping.
+
+    This class is a custom routing converter defining the 'PID' type.
+    See http://werkzeug.pocoo.org/docs/0.12/routing/#custom-converters.
+
+    Use ``pid`` as a type in the route pattern, e.g.: the use of
+    route decorator: ``@blueprint.route('/record/<pid:pid_value>')``,
```
slint: The converter requires at least one parameter (the pid_type). Also using the extra `getter` and `record_class` aguments is common. [Example from Zenodo](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/views.py#L123-L127):

```python
@blueprint.route('/records/<pid(recid,record_class="zenodo.modules.records.api:ZenodoRecord"):pid_value>')
```

PR #168, opened 2017-11-23
views: configurable error handling

### LINKED ISSUE #152: views: PIDConverter prevents from sending additional data during a GET of a deleted record
**Problem:**
In B2SHARE we would like to return the reason why a record was deleted when a client makes a GET request on a deleted request.
This is currently not possible as the [`PIDConverter`](https://github.com/inveniosoftware/invenio-records-rest/blob/d36b628c5687ca7342ce756bf42da53befb18eb7/invenio_records_rest/utils.py#L189) will [raise a `PIDDeletedRESTError` exception](https://github.com/inveniosoftware/invenio-records-rest/blob/d36b628c5687ca7342ce756bf42da53befb18eb7/invenio_records_rest/utils.py#L156) which bypasses the serializer.

I don't have yet any clear solution for this problem.
This is not an urgent issue.

slint:
* Introduces the 'error_handlers' key for the 'RECORDS_REST_ENDPOINTS'
  config variable which allows for registration of custom error
  uandlers for exceptions and HTTP status codes occuring in specific
  endpoint of the REST API blueprint views. (closes #152)

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -28,6 +28,7 @@ from __future__ import absolute_import, print_function
 
 import copy
 import uuid
+from collections import defaultdict
 from functools import partial, wraps
 
 from elasticsearch.exceptions import RequestError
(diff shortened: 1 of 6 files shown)


invenio_records_rest/views.py
```diff
@@ -103,10 +117,16 @@ def create_blueprint(endpoints):
         url_prefix='',
     )
 
+    error_handlers_registry = defaultdict(dict)
     for endpoint, options in (endpoints or {}).items():
+        error_handlers = options.pop('error_handlers', None)
```
krzysztof: could this be ``options.pop('error_handlers', {})`` instead? If so, remove ``(error_handlers or {})`` below.

invenio_records_rest/views.py
```diff
@@ -88,6 +89,19 @@ def elasticsearch_badrequest_error(error):
                 return handler(error)
         return error
 
+    for exception_cls, handlers in (error_handlers_registry or {}).items():
```
krzysztof: Maybe fix the default for the argument at the top of the function? ``error_handlers_registry = error_handlers_registry or {}``

krzysztof:
LGTM otherwise.

invenio_records_rest/views.py
```diff
@@ -97,16 +131,24 @@ def create_blueprint(endpoints):
     :params endpoints: Dictionary representing the endpoints configuration.
     :returns: Configured blueprint.
     """
+    endpoint = endpoints or {}
```
krzysztof: ``endpointS`` ?

Issue #172, opened 2017-11-27
views: indexing


dinosk:
Currently there is no Elasticsearch indexing for the actions performed from the views.
It can be added after the `db.session.commit()` so that nothing is indexed after a potential db rollback.

PR #174, opened 2017-11-27
global: indexing after actions

dinosk:
* CHANGE indexing of changes after PUT, DELETE, PATCH, POST

* Extend the tests to check for proper indexing
   Closes https://github.com/inveniosoftware/invenio-records-rest/issues/172

Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>
Co-authored-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/invenio_records_rest/config.py
+++ b/invenio_records_rest/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2016 CERN.
+# Copyright (C) 2016, 2017 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
@@ -27,6 +27,7 @@
 from __future__ import absolute_import, print_function
 
 from flask import request
+from invenio_indexer.api import RecordIndexer
 from invenio_search import RecordsSearch
 
 from .facets import terms_filter
@@ -43,6 +44,7 @@ RECORDS_REST_ENDPOINTS = dict(
         pid_minter='recid',
         pid_fetcher='recid',
         search_class=RecordsSearch,
+        indexer_class=RecordIndexer,
         search_index=None,
         search_type=None,
         record_serializers={
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -37,6 +37,7 @@ from flask import Blueprint, abort, current_app, jsonify, make_response, \
 from flask.views import MethodView
 from flask_babelex import gettext as _
 from invenio_db import db
+from invenio_indexer.api import RecordIndexer
 from invenio_pidstore import current_pidstore
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_pidstore.resolver import Resolver
@@ -163,6 +164,7 @@ def create_url_rules(endpoint, list_route=None, item_route=None,
                      record_serializers_aliases=None,
                      record_loaders=None,
                      search_class=None,
+                     indexer_class=None,
                      search_serializers=None,
                      search_index=None, search_type=None,
                      default_media_type=None,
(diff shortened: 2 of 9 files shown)


tests/test_views_item_patch.py
```diff
@@ -57,34 +58,47 @@ def test_valid_patch(app, test_records, test_patch, content_type):
         assert res.status_code == 200
 
         # Check that year changed.
-        assert previous_year != get_json(client.get(url))['metadata']['year']
+        new_year = get_json(client.get(url))['metadata']['year']
+        assert previous_year != new_year
+        es.indices.flush('*')
```
diegodelemos: Should we use [`flush_and_refresh`](https://github.com/inveniosoftware/invenio-search/blob/ebe51c79a7dde1b0b2f491935f7a4339f91c4282/invenio_search/ext.py#L176) instead?  since it is taking care of more than `flush` and also would wait if indices could not be indexed straight away.

dinosk: Good idea

invenio_records_rest/config.py
```diff
@@ -43,6 +44,7 @@ def _(x):
         pid_minter='recid',
         pid_fetcher='recid',
         search_class=RecordsSearch,
+        indexer_class=RecordIndexer,
```
lnielsen: I think likely this should be an object instance and not a class. This will allow to customize the indexer with the parameters needed (e.g. specific queue names etc)

invenio_records_rest/views.py
```diff
@@ -564,6 +575,10 @@ def post(self, **kwargs):
 
         db.session.commit()
 
+        # Index the record
+        if self.indexer_class:
+            self.indexer_class().index_by_id(record_uuid)
```
lnielsen: I think it's better to use ``self.indexer.index(record)`` because ``index_by_id()`` will make a query to the database which is not needed, since you already have the record available.

invenio_records_rest/config.py
dinosk: Right, it's changed to indexer and we had to pass `indexer=lambda: RecordIndexer()` because of this line in conftest: `RECORDS_REST_ENDPOINTS=copy.deepcopy(config.RECORDS_REST_ENDPOINTS)` which will complain about the missing application context.

invenio_records_rest/config.py
```diff
@@ -43,6 +44,7 @@ def _(x):
         pid_minter='recid',
         pid_fetcher='recid',
         search_class=RecordsSearch,
+        indexer=lambda: RecordIndexer(),
```
lnielsen: Why the lambda? Is there problems if it would be just ``RecordIndexer()``?

dinosk: Yes, without it it complained that the application context was missing when building the app fixture for the tests because of this deepcopy: https://github.com/inveniosoftware/invenio-records-rest/blob/master/tests/conftest.py#L138
I realise that the copying was required for testing with different configurations though, so I'm not sure if the best thing would be to remove the lambda, pass directly the object instance and rewrite the tests.

nharraud: `invenio-indexer` CLI currently prevents any customization of the indexer class. It creates the default RecordIndexer class when doing a reindex.

Thus we have to make a choice:
1/ Not provide a way to override the RecordIndexer, only use the default class.
2/ Enable invenio-indexer CLI to use another class.

If we stay with the currently suggested solution developers will try to change the class and have a surprise when they do `invenio index reindex`.

tests/test_views_item_put.py
```diff
@@ -83,26 +88,37 @@ def test_valid_put_etag(app, test_records, content_type):
...
     with app.test_client() as client:
-        url = record_url(pid)
+        HEADERS = [
+            ('Accept', 'application/json'),
+            ('Content-Type', content_type)
+        ]
+        HEADERS.append(('Content-Type', content_type))
+
+        # Create record
+        res = client.post(
```
nharraud: why are we not using the `indexed_records` fixture?

invenio_records_rest/config.py
lnielsen: Ok, then we can put it back to ``indexer_class = RecordIndexer`` - sorry for the noise

tests/test_views_item_put.py
dinosk: We would have 3 indexed records remaining and one of them could be returned when we search for it after deleting. It's also cleaner to have 1 record, delete it and then have 0.

PR #184, opened 2017-12-06
Fix wrong default indexer

ntarocco:


invenio_records_rest/views.py
```diff
@@ -257,7 +257,7 @@ def links_factory(pid, record=None, **kwargs):
         search_class, default=RecordsSearch
     )
     indexer_class = obj_or_import_string(
-        indexer_class, default=RecordIndexer
+        indexer_class, default=None
```
lnielsen: Is this because you have problems in with deposit? It ought to have set this to None in Invenio-Deposit.

lnielsen: Ah, is not fixed in Invenio-Deposit....so I think this fix should be applied to Invenio-Deposit instead....I'll pass by the office.

dinosk: @lnielsen I think this is a mistake I did, we had set it to None when we were using the object instance `indexer` and testing with invenio-deposit to avoid the double indexing, but I refactored it and set the default to RecordIndexer.

diegodelemos: We decided to isolate it here, setting it to none by default, so no new PR on Deposit needed to be done back then

lnielsen: But shouldn't we enable indexing by default? and make fix to Invenio-Deposit?

ntarocco:
@lnielsen @dinosk I applied the fix to the param set to `RecordIndexer`, have a look if you see anything else!

PR #185, opened 2017-12-11
views: support for sub-records

jirikuncar:
* BETTER Adds support for sub-records (parent/child) under the same
  url structure `/records/1/subrecord/2`.

* FIX Allows passing `LazyPIDValue` as argument to `url_for`.


invenio_records_rest/links.py
```diff
@@ -2,6 +2,9 @@
 #
 # This file is part of Invenio.
 # Copyright (C) 2015, 2016 CERN.
+# Copyright (C) 2017 Swiss Data Science Center (SDSC)
```
lnielsen: FYI: We will wait with the integration until we've implemented the license change of the repository GPL to MIT to avoid further complicating the license change (basically integrating now, would mean we would have to ask SDSC for permission for the change from GPL to MIT). Sorry for this, we will try to attack this repo as one of the first to speed it up.

invenio_records_rest/links.py
```diff
@@ -2,6 +2,9 @@
...
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
```
lnielsen: I don't think these two extra lines should be added. The copyright should just state who the copyright holder is (must be a legal entity) - e.g. either it's SDSC (if SDSC is a legal entity) or it is EPFL + ETHZ (if SDSC is not a legal entity)?
For crediting we can add in AUTHORS.rst that SDSC is a partnership between EPFL/ETHZ.

invenio_records_rest/links.py
jirikuncar: I can send you written permission if you need one.

invenio_records_rest/links.py
jirikuncar: @lnielsen this is the official way for now according to EPFL/ETH layers for now.

invenio_records_rest/links.py
lnielsen: The problem is fully on the CERN-end. We're awaiting answer from CERN KT on how we need to obtain permissions from third-party in a legally correct way. I'll try to make sure the PR goes in before the CERN annual closure.

invenio_records_rest/links.py
jirikuncar: The problem is that both institutions need to be mentioned in full name. It should be easier with the NOTICE file when the MIT license is applied.

lnielsen:
Closing PR because it's been stale for a long time.

PR #186, opened 2018-02-04
global: fix build

jacquerie:
Our build appears to be failing because of an incompatibility between `invenio-records-rest` and the latest beta release of `marshmallow`. See: https://travis-ci.org/inspirehep/inspire-docker/jobs/337193978#L7402.

I'm already seeing this failure locally on just `invenio-records-rest`, let's see if Travis agrees with me...


jacquerie:
> let's see if Travis agrees with me...

It doesn't. I guess the problem is that we install dependencies with `--pre`, so we pick `marshmallow==3.0.0b7` instead of `marshmallow==2.15.0`. Therefore the pin should be on our side.

dfdan:
a little late, but for info mashmallow==3.0.0b4 works fine for me - 3.0.0b7 which I pulled in today caused errors. I guess if it works with 2.15 that's more stable, but this might help pin down where the problem has snuck in...

PR #205, opened 2018-05-14
serializers: drop usage of request.is_xhr for JSON

### LINKED ISSUE #204: Deprecation warning for "request.is_xhr"
[`request.is_xhr`](https://github.com/inveniosoftware/invenio-records-rest/blob/97c320c79e7c609b333a2a73fdfec3a26834dff3/invenio_records_rest/serializers/json.py#L29) has been deprecated in [`werkzeug>=0.13`](https://github.com/pallets/werkzeug/blob/1c70c3e1855d32a285843c0851f52e20160f6656/werkzeug/wrappers.py#L694)

Some further explanation for why this is happening: https://github.com/pallets/werkzeug/issues/1077

In our case, this is only used to check for pretty-printing JSON... I think we should probably change it to what [flask decided to change it to as well](https://github.com/pallets/flask/pull/2193/files#diff-dc6f589d1b53806b6a6287c182b461b2L251), i.e. `s/request.is_xhr/current_app.debug`

slint:
* Drops usage of `request.is_xhr` since it's deprecated in
  flask/werkzeug. (closes #204)
* Prints pretty JSON in debug mode.

### DIFF
--- a/invenio_records_rest/serializers/json.py
+++ b/invenio_records_rest/serializers/json.py
@@ -22,15 +22,7 @@ class JSONSerializerMixin(SerializerMixinInterface):
     @staticmethod
     def _format_args():
         """Get JSON dump indentation and separates."""
-        # Ensure we can run outside a application/request context.
-        try:
-            pretty_format = \
-                current_app.config['JSONIFY_PRETTYPRINT_REGULAR'] and \
-                not request.is_xhr
-        except RuntimeError:
-            pretty_format = False
-
-        if pretty_format:
+        if request and request.args.get('prettyprint'):
             return dict(
                 indent=2,
                 separators=(', ', ': '),
--- a/tests/test_serializer_json.py
+++ b/tests/test_serializer_json.py
@@ -78,12 +78,10 @@ def test_serialize_pretty(app):
     pid = PersistentIdentifier(pid_type='recid', pid_value='2'),
     rec = Record({'title': 'test'})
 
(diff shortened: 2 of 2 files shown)


invenio_records_rest/serializers/json.py
```diff
@@ -25,8 +25,8 @@ def _format_args():
         # Ensure we can run outside a application/request context.
         try:
             pretty_format = \
-                current_app.config['JSONIFY_PRETTYPRINT_REGULAR'] and \
-                not request.is_xhr
+                current_app.config['JSONIFY_PRETTYPRINT_REGULAR'] or \
```
lnielsen: I think by default we should not prettyprint the JSON (to avoid adding lots of spaces/tabs into the output). Perhaps, it's better if we just add a request parameter that you can use instead.....e.g. ``/api/records/?prettyprint``

slint: The default value for the config was changed to `False` in Flask v0.12, (so no prettyprint by default). So in the end, which of the following should we use?

- [ ] `JSONIFY_PRETTYPRINT_REGULAR`
- [ ] `current_app.debug`
- [ ] `'prettyprint' in request.args`

lnielsen: ``if request and request.args.get('prettyprint')``? 

I think it's still useful to have the JSON formatted nicely for display in a browser.....

lnielsen: so simply remove debug and ``JSONIFY_PRETTYPRINT_REGULAR``

PR #206, opened 2018-05-15
serializers: addition of generic fields

dinosk:
Addresses https://github.com/inveniosoftware/cookiecutter-invenio-datamodel/issues/4

### DIFF
--- a/tests/test_custom_fields.py
+++ b/tests/test_custom_fields.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio custom schema fields tests."""
+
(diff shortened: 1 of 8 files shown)


lnielsen:
Can I ask you to add some tests to avoid lowering the code coverge? I'm not sure there's tests elsewhere already written - but e.g. here you can find some problematic unicode characters: https://github.com/zenodo/zenodo/blob/master/tests/unit/records/test_schemas_legacyjson_load.py#L826-L836

tests/test_custom_fields.py
```diff
@@ -0,0 +1,43 @@
...
+
+    date_string_field = DateString(attribute='metadata.datestring')
+    sanitized_html_field = SanitizedHTML(attribute='metadata.sanitized_html')
+    sanitized_unicode_field = SanitizedUnicode(
+        attribute='metadata.sanitized_unicode_field')
+    trimmed_string_field = TrimmedString(
+        attribute='metadata.trimmed_string_field')
+
+
+def test_load_custom_fields(app):
```
dinosk: This test case checks if the fields work when loading data

PR #207, opened 2018-05-16
* FIX test_error_handlers invalid-json test

ppanero:
The ``invalid-json`` test in test_error_handlers was comparing to the wrong string (maybe coming from an old version). After testing with a simple POST and PUT with ``'{invalid-json' as data, it can be seen that the string returned by the server is:

```'Failed to decode JSON object: Expecting property name  enclosed in double quotes: line 1 column 2 (char 1)```


lnielsen:
This could be related to Flask 0.12 vs 1.0 differences.

ppanero:
@lnielsen it was actually that the message changes if you are in debug mode (and I was). Closing PR

PR #208, opened 2018-05-23
serializers: marshmallow loader

dinosk:
Addresses https://github.com/inveniosoftware/cookiecutter-invenio-datamodel/issues/19
Depends on https://github.com/inveniosoftware/cookiecutter-invenio-datamodel/pull/27


tests/test_marshmallow_loader.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio marshmallow loader tests."""
+
+from __future__ import absolute_import, print_function
+
+# from invenio_pidstore.models import PersistentIdentifier
+# from invenio_records import Record
```
switowski: I think those were forgotten to be removed :P

dinosk: mah gad, jesus take the wheel :D
fixing, thanks

invenio_records_rest/serializers/loaders/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
lnielsen: I think this should go to ``invenio_records_rest/loaders/__init__.py`` similar to how it was in the data model package.

invenio_records_rest/serializers/loaders/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Loaders."""
+
+from .marshmallow import json_patch_loader, marshmallow_loader
+from ..schemas.json import RecordSchemaJSONV1
+
+json_v1 = marshmallow_loader(RecordSchemaJSONV1)
```
lnielsen: Should they actually be here (same goes for ``invenio_records_rest/serializers:json_…``)? I think we originally added them in ``invenio_records_rest/serializers:json_`` because we thought we could reuse them with the need for further schema, but that's no longer the case as far as I'm aware?

switowski:
That should not be closed :/ I did a bad force push

PR #210, opened 2018-05-23
global: load blueprints from entrypoint

### LINKED ISSUE #209: global: load blueprints via "invenio_base.api_blueprints"
Based on the changes introduced in inveniosoftware/invenio-base#133, it will be possible to load blueprints via a function, allowing for the application to have loaded any `RECORDS_REST_ENDPOINTS` config from other extensions before creating and registering the [`invenio_records_rest` blueprint](https://github.com/inveniosoftware/invenio-records-rest/blob/7c32876db1af213e81cbcaa7227eccb64f3c83b2/invenio_records_rest/ext.py#L90-L92). This allows for flexible "feature-flags" as described in inveniosoftware/cookiecutter-invenio-datamodel#18.

slint:
* Introduces an "invenio_base.api_blueprints" to load REST API
  blueprints via a function. (closes #209)

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -102,6 +102,7 @@ from invenio_records_rest import InvenioRecordsREST
 from invenio_records_rest.config import RECORDS_REST_ENDPOINTS
 from invenio_records_rest.facets import terms_filter
 from invenio_records_rest.utils import PIDConverter
+from invenio_records_rest.views import create_blueprint_from_app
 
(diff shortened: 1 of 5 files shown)


slint:
Depends on inveniosoftware/invenio-base#133. 

Tested on master branch of Zenodo successfully, without requiring any changes (ie. this won't break everything)

PR #211, opened 2018-05-25
serializers: addition of marshmallow loaders

dinosk:
Addresses https://github.com/inveniosoftware/cookiecutter-invenio-datamodel/issues/19


switowski:
@lnielsen is the current structure of this module something that you had in mind (based on the comments from https://github.com/inveniosoftware/invenio-records-rest/pull/208)?

PR #212, opened 2018-05-25
serializers: PID in marshmallow context

slint:
Depends on #211


slint:
@lnielsen WDYT? This solves the issue where the `id` is not picked up by the serializer for the response.

PR #215, opened 2018-06-07
Component: authentication on resource list get action

ppanero:
Make the GET operation over the RecordsListResource require authentication, under ``allow_all` by default to maintain the current behaviour.

Signed-off-by: Pablo Panero <pablo.panero@cern.ch>


ppanero:
Tests are failing due to:
```
self = <sqlalchemy.engine.base.Connection object at 0x7f284429bcd0>
    def _revalidate_connection(self):
        if self.__branch_from:
            return self.__branch_from._revalidate_connection()
        if self.__can_reconnect and self.__invalid:
            if self.__transaction is not None:
                raise exc.InvalidRequestError(
                    "Can't reconnect until invalid "
                    "transaction is rolled back")
            self.__connection = self.engine.raw_connection(_connection=self)
            self.__invalid = False
            return self.__connection
>       raise exc.ResourceClosedError("This Connection is closed")
``` 
Known issue (https://github.com/kvesteri/sqlalchemy-continuum/issues/188), need to pin ``'SQLAlchemy-Continuum==1.3.4'``, but did not want to enforce it on this PR.

ppanero:
Tests should pass when https://github.com/inveniosoftware/invenio-db/pull/111 is merged

invenio_records_rest/config.py
```diff
@@ -96,6 +96,7 @@ def deleted_pid_error_handler(error):
             'pid_fetcher': '<registered-pid-fetcher>',
             'pid_minter': '<registered-minter-name>',
             'pid_type': '<record-pid-type>',
+            'read_list_permission_factory_imp': permission_check_factory(),
```
lnielsen: just seeing this now - is it not better named ``list_permission_factory``? create, read, update, delete, list

ppanero: Yeah maybe, if post_list is create, maybe this should be list. Should I do another PR? Sorry for the inconvenience :/

ntarocco: @lnielsen yes you are right, for me it was clear enough, but it is indeed more consistent with the rest of the code. We can see how to proceed.

PR #224, opened 2018-08-15
schemas: addition of PersistentIdentifier field

dinosk:
Addresses https://github.com/inveniosoftware/cookiecutter-invenio-datamodel/issues/30

### DIFF
--- a/tests/test_serializer_json.py
+++ b/tests/test_serializer_json.py
@@ -16,6 +16,8 @@ from invenio_pidstore.models import PersistentIdentifier
 from invenio_records import Record
 from marshmallow import Schema, fields
 
+from invenio_records_rest.schemas.fields import \
+    PersistentIdentifier as PIDField
 from invenio_records_rest.serializers.json import JSONSerializer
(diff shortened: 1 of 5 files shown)


lnielsen:
We need to add some tests.

tests/test_serializer_json.py
```diff
@@ -16,14 +16,16 @@
...
+    PersistentIdentifier as PIDField
 from invenio_records_rest.serializers.json import JSONSerializer
 
 
 def test_serialize():
     """Test JSON serialize."""
     class TestSchema(Schema):
         title = fields.Str(attribute='metadata.mytitle')
-        id = fields.Str(attribute='pid.pid_value')
+        id = PIDField(attribute='pid.pid_value')
```
ChiaraBi: Shouldn't it be like this even [here](https://github.com/inveniosoftware/invenio-records-rest/pull/224/files#diff-e4dfb0584ad0d70acf59d636a0c8ff24R42)?

PR #225, opened 2018-08-16
schemas: Nested class w/out recursive error check

### LINKED ISSUE #223: Change validation error message when creating a record via the API
Trying to create a test record I saw the following message:
```
$ curl -k -H "Content-type: application/json" -XPOST https://localhost:5000/api/records/ -d '{"title":"aaa1", "name":"test"}'
{"status": 400, "message": "Validation error.", "errors": [{"field": "name", "message": "Unknown field name name"}, {"field": "name", "message": "Missing data for required field."}]}%      
```
which doesn't mention that the required `name` field is inside the `contributors` field.
```
$ curl -k -H "Content-type: application/json" -XPOST https://localhost:5000/api/records/ -d '{"title":"aaa1", "contributors":[{"name": "Doe, John"}]}'
{"created"... works ok 
```

dinosk:
* ADD Nested class which overrides the _validate_missing method,
  and reports errors for the topmost missing fields.

Closes #223 

### DIFF
--- a/tests/test_marshmallow_loader.py
+++ b/tests/test_marshmallow_loader.py
@@ -17,15 +17,27 @@ from helpers import get_json
 from marshmallow import Schema, fields
 
 from invenio_records_rest import loaders
-from invenio_records_rest.schemas import RecordSchemaJSONV1
+from invenio_records_rest.schemas import Nested, RecordSchemaJSONV1
 
 
 class _TestSchema(Schema):
-        """Test schema."""
+    """Test schema."""
 
-        title = fields.Str(required=True, attribute='metadata.mytitle')
-        random = fields.Str(required=True, attribute='metadata.nonexistant')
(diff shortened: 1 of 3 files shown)


tests/test_marshmallow_loader.py
```diff
@@ -86,6 +98,31 @@ def test_marshmallow_load_errors(app, db, es, test_data, search_url,
...
+    }
+
+    with app.test_client() as client:
+        HEADERS = [
+            ('Accept', 'application/json'),
+            ('Content-Type', 'application/json')
+        ]
+
+        # Create record
+        missing_top_most_required_field = dict(nested="test")
```
diegodelemos: Just as a useless comment, I would use  `{'nested': 'test'}` to make it more readable as JSON is our internal representation, however, I see the dict constructor is used everywhere...

PR #226, opened 2018-08-20
views: suggesters support source filtering for ES5

iulianav:
Make ES5 suggesters support [source filtering](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-request-source-filtering.html) in order to save network overhead by filtering out useless fields from the `_source` key (or just strip `_source` altogether) returned in the response.
Suggester fields configuration should now contain a `_source` key that will specify the source filtering for each suggestion field. The possible values for the `_source` key are the same as the ones specified in source filtering documentation.

Example configuration:
```
'suggesters': {
        'author': {
            '_source': ['name', 'age'],
            'completion': {
                'field': 'author_suggest',
            },
        }
}
```

With the current test structure, it is not possible to add a reliable test to handle both ES2 and ES5.

Signed-off-by: Iuliana Voinea <iulianavoinea96@gmail.com>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.2.1 (released 2018-09-17)
+
+- Adds source filtering support for ES 5.
+
 Version 1.2.0 (released 2018-08-24)
 
 - Adds PersistentIdentifier field to handle record PIDs.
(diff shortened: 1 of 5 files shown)


iulianav:
@lnielsen as I mentioned, with the current test structure, it am not sure it is possible to add a reliable test to handle both ES2 and ES5. What do you think?

lnielsen:
Sorry, didn't see the last comment. In the test I think you can do the same as the in the code 

```python 
if ES_VERSION[0] >= 5:
    # assert that a non-requested field *is not* returned
else: 
    # assert that a non-requested field *is* returned
```

This way Travis will run both on v2 and v5 and the test coverage should make sure we have both lines covered.

drjova:
ping @lnielsen

PR #230, opened 2018-11-15
facets: change `str` to `text_type`

pazembrz:
Fixes utf in types for ES filter.


slint:
@ntarocco I think you also had this bug on CDS Videos?

egabancho:
The one I remember was with the sorting https://github.com/inveniosoftware/invenio-records-rest/commit/83f802f933b127ba282588f9a90d9dbfb01883c4, but yes, essentially the same issue.

Issue #232, opened 2018-11-16
fields: PersistentIdentifier not (de)serializing as expected


slint:
The [`.schemas.fields.PersistentIdentifier`](https://github.com/inveniosoftware/invenio-records-rest/blob/ada7d4d556d7b92cab88039bbfa87e4781e320ab/invenio_records_rest/schemas/fields/persistentidentifier.py#L20-L26) marshmallow field, is assuming a wrong method signature when overriding [`Field._deserialize`](https://marshmallow.readthedocs.io/en/2.x-line/api_reference.html#marshmallow.fields.Field._deserialize) and `Field._serialize`: 

```python
# wrong
...
def _deserialize(self, value, attr, context):
    # "context" is actually the original "data" passed to "Schema(...).load"
    # and not the "Schema(context=context).load(...)" (see docs link above)
    pid = context.get('pid')
    return pid.get('pid_value') if pid else missing
...

# correct
def _deserialize(self, value, attr, data):
    pid = self.context.get('pid')
    return pid.get('pid_value') if pid else missing
```

When addressing this, keep in mind that:

- When doing a `POST /api/records`, there will be no [`request.view_args['pid_value']`](https://github.com/inveniosoftware/invenio-records-rest/blob/ada7d4d556d7b92cab88039bbfa87e4781e320ab/invenio_records_rest/loaders/marshmallow.py#L87-L90) and thus it won't be available in the loader schema's context, since the route doesn't have the PID yet, opposed to e.g. `PUT /api/records/<pid(recid):pid_value>`. The PID [will be minted](https://github.com/inveniosoftware/invenio-records-rest/blob/ada7d4d556d7b92cab88039bbfa87e4781e320ab/invenio_records_rest/views.py#L596) though and [passed to the serializer](https://github.com/inveniosoftware/invenio-records-rest/blob/ada7d4d556d7b92cab88039bbfa87e4781e320ab/invenio_records_rest/views.py#L606-L607) and thus will be available in the context of `Field._serialize` [because of this guy](https://github.com/inveniosoftware/invenio-records-rest/blob/ada7d4d556d7b92cab88039bbfa87e4781e320ab/invenio_records_rest/serializers/marshmallow.py#L32-L33).
[...]

PR #234, opened 2018-11-26
Marshmallow schema and field updates for PID and other helpers

ppanero:
Fixes: https://github.com/inveniosoftware/invenio-records-rest/issues/232
Fixes: [cookiecutter-invenio-instance#97](https://github.com/inveniosoftware/cookiecutter-invenio-instance#97)
In order to use it the loader should be overwritten with:

```python
record_loaders={
            'application/json': ('invenio_records_rest.loaders:json_pid_checker')
        },
```

And add the PID attribute name on the environment / config file:

```python
[ INVENIO_ ]RECORDS_REST_PID_ATTR_NAME = 'control_number'
```

For the code:

- Should the "get_pid" function be customizable (e.g. through ``import_string()``)? If not, is ``json.py`` the best place to be for this function, ``utils.py``?
- There is a test file for [custom marshmallow loaders](https://github.com/inveniosoftware/invenio-records-rest/blob/master/tests/test_marshmallow_loader.py). Should it be overwritten? Should I create a new one? Mainly to comply with coveralls.

For the tests:

- Create record (post), chek record PID.
- Update with wrong PID, check it does not change (search + DB)
- Update with no PID, check it does not change (search + DB)

Small punctualization: In order to follow the naming convention of this repo/other invenio repos, shouldn't the file be named ``test_loader_marshmallow`` instead of ``test_marshmallow_loader`` (e.g. check the serializers).

### DIFF
--- a/invenio_records_rest/loaders/__init__.py
+++ b/invenio_records_rest/loaders/__init__.py
@@ -9,7 +9,7 @@
 """Loaders for deserializing records in the REST API."""
 
 from .marshmallow import json_patch_loader, marshmallow_loader
-from ..schemas import RecordSchemaJSONV1
+from ..schemas import RecordSchemaJSONV1, RecordMetadataSchemaJSONV1
 
 json_v1 = marshmallow_loader(RecordSchemaJSONV1)
 """Simple example loader that will take any JSON."""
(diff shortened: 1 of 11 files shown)


ppanero:
@slint The changes are missing the tests:
- Test the new ``RecordMetadataSchemaJSONV1`` in order to check proper ``RECID_FIELD`` value injection.
- Cross check the ``test_marshmallow_load`` in ``test_marshmallow_loader.py``. Since the serializer is not declared the following assertion will never be correct:
```python
        assert 'pid' in data_dump.data.get('metadata')
```
- Create a test to check the serialization with the ``PersistentIdentifier`` field. 

Lastly and for consistency the name of ````test_marshmallow_loader.py`` should be changed to ``test_loader_marshmallow.py``.

ppanero:
@slint It has been merged, I am going to test it. Just that I cannot put myself as a reviewer :rofl:

ppanero:
Tested. Also closes https://github.com/inveniosoftware/cookiecutter-invenio-instance/issues/97

PR #239, opened 2018-12-12
error handling: return serializable unhandled error

ppanero:
* Adds an exception that inherits from  REST Exception and represents an unhandled exception.
* Logs the exception so it goes to Sentry/Logging.
* Returns a serializable response so the App does not crash (Internal Server Error).
* Users should change ``RECORDS_REST_ELASTICSEARCH_ERROR_HANDLERS`` to adapt the errors to their needs (e.g. return parsing errors, which at the moment are not handled).
Closes https://github.com/inveniosoftware/invenio-indexer/issues/88
Outdates https://github.com/inveniosoftware/troubleshooting/issues/24

### DIFF
--- a/invenio_records_rest/errors.py
+++ b/invenio_records_rest/errors.py
@@ -187,3 +187,11 @@ class JSONSchemaValidationError(RESTValidationError):
         super(RESTValidationError, self).__init__(**kwargs)
         self.description = 'Validation error: {0}.'.format(
             error.message if error else '')
+
+
+class UnhandledElasticsearchError(RESTException):
+    """Failed to handle exception."""
+
+    code = 500
+    description = 'An internal server error occurred when handling the ' \
+                  'request.'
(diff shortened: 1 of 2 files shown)


invenio_records_rest/views.py
```diff
@@ -95,7 +95,10 @@ def elasticsearch_badrequest_error(error):
         for cause_type, handler in handlers.items():
             if cause_type in cause_types:
                 return handler(error)
-        return error
+
+        exception = UnhandledElasticsearchError()
```
lnielsen: It seems like the Elasticsearch error is swallowed here thus you won't know exactly what error couldn't be handled.

invenio_records_rest/errors.py
```diff
@@ -187,3 +187,10 @@ def __init__(self, error=None, **kwargs):
         super(RESTValidationError, self).__init__(**kwargs)
         self.description = 'Validation error: {0}.'.format(
             error.message if error else '')
+
+
+class UnhandledElasticsearchError(RESTException):
+    """Failed to handle exception."""
+
+    code = 400
+    description = 'An exception occured but there was no handler for it.'
```
lnielsen: This is displayed to users, so I would make it a bit more user-friendly in the wording. Is it correct that this error happens when a we execute a bad search query? If so, I would default to something like: ```Bad query syntax.```.

Alternatively, if the error is on our side, I would return a 500 error, with something like ``An internal server error happened during the search.``


```Failed to run

PR #240, opened 2018-12-12
docs: query parsing example code

### LINKED ISSUE #237: docs: document customisation of search factory and query parser
Is it possible to modify the `querystring` from the flask api? For example if the incoming request is `http://localhost:5001/records/?q=abc%20cde` I want to update the `querystring` value from "abc cde" to "\*abc\* \*cde\*"

kprzerwa:
(closes #237 )

### DIFF
--- a/invenio_records_rest/__init__.py
+++ b/invenio_records_rest/__init__.py
@@ -316,8 +316,28 @@ Its features include:
 - boosting
 
 
-A custom query parser can also be plugged in by setting the ``search_factory``
-option in ``RECORDS_REST_ENDPOINTS`` to point to the function implementing it.
+A custom query parser can also be plugged in by setting the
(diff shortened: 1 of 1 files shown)


lnielsen:
Failing because the code is written as a doctest. Probably the best is just to change to something like:

```rst
.. code-block:: python

    # config.py
    ...
```

kprzerwa:
@lnielsen actually it was just not properly formatted in the doctest

invenio_records_rest/__init__.py
```diff
@@ -317,7 +317,26 @@
 
 
 A custom query parser can also be plugged in by setting the ``search_factory``
```
ppanero: Is not on the changes of the commit but the exact name of the setting is ``search_factory_imp`` (is missing the "_imp" part)

PR #242, opened 2019-01-21
WIP: views: allow ES indexing parameters from POST request

ppanero:
Fixes https://github.com/inveniosoftware/invenio-records-files/issues/53 (Probably worth it to change or duplicate the issue here).

It allows binary files content extraction by using Elasticsearch attachment pipeline. In order to do so, it has to send the parameter ``pipeline`` with a value equals to the pipeline id (ej. ``attachment``).

Needs: https://github.com/inveniosoftware/invenio-indexer/pull/99.

### DIFF
--- a/invenio_records_rest/utils.py
+++ b/invenio_records_rest/utils.py
@@ -112,6 +112,25 @@ def check_elasticsearch(record, *args, **kwargs):
     return type('CheckES', (), {'can': can})()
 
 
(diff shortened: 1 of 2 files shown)


ppanero:
Decided to be done via signals and new methods directly in invenio-indexer. Closing in favor of: https://github.com/inveniosoftware/invenio-indexer/pull/99

PR #249, opened 2019-03-28
config: addition of configurable search results size

drjova:
* Adds ``RECORDS_REST_DEFAULT_SEARCH_RESULTS_SIZE``
  variable to change the default ``size`` of the search results.
  The default value remains ``10``.

Signed-off-by: Harris Tzovanakis <me@drjova.com>

### DIFF
--- a/invenio_records_rest/config.py
+++ b/invenio_records_rest/config.py
@@ -363,3 +363,6 @@ RECORDS_REST_ELASTICSEARCH_ERROR_HANDLERS = {
     ),
 }
 """Handlers for ElasticSearch error codes."""
+
+RECORDS_REST_DEFAULT_RESULTS_SIZE = 10
+"""Default search results size."""
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,7 @@ setup_requires = [
 
 install_requires = [
     'arrow>=0.12.1',
+    'attrs>=17.4.0',
     'bleach>=2.1.3',
     'ftfy>=4.4.3,<5.0',
     'Flask>=0.11.1',
@@ -84,7 +85,7 @@ install_requires = [
     'invenio-indexer>=1.0.0',
     'marshmallow>=2.5.0',
     'python-dateutil>=2.4.2',
-    'six>=1.10',
+    'six>=1.11',
     'webargs>=1.3.2',
 ]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@
 Changes
 =======
 
+Version 1.4.1 (released 2019-04-02)
+
+- Added ``RECORDS_REST_DEFAULT_RESULTS_SIZE`` variable to change the default
+  ``size`` of the search results. The default value remains ``10``.
+
 Version 1.4.0 (released 2019-02-22)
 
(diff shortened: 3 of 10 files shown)


invenio_records_rest/config.py
```diff
@@ -363,3 +363,6 @@ def deleted_pid_error_handler(error):
     ),
 }
 """Handlers for ElasticSearch error codes."""
+
+RECORDS_REST_DEFAULT_SEARCH_RESULTS_SIZE = 10
+"""Default search results size.""""
```
egabancho: Should we, perhaps, move it to `RECORDS_REST_ENDPOINTS` like for example `max_result_window`? (I am really just asking)

drjova: 👍 don't have strong opinion, we'll add yet another rule in ``create_url_rules`` (it's quite big). The main idea behind this PR is to have consistency in all of your searches, without add the logic everywhere in the UI. What I mean is, having separate search results size per type it's not the real case, but we can do it. ☺️ Do you think will be useful per type?

lnielsen:
It probably makes sense to be able to configure it per endpoint, but it's anyway hardcoded today, so enabling to configure is also fine by me. I.e. either of them works for me.

invenio_records_rest/config.py
```diff
@@ -363,3 +363,6 @@ def deleted_pid_error_handler(error):
     ),
 }
 """Handlers for ElasticSearch error codes."""
+
+RECORDS_REST_DEFAULT_SEARCH_RESULTS_SIZE = 10
```
lnielsen: ``RECORDS_REST_DEFAULT_SEARCH_RESULTS_SIZE`` -> 
``RECORDS_REST_DEFAULT_RESULTS_SIZE``? 

To make it a bit shorter?

setup.py
```diff
@@ -74,6 +74,7 @@
 
 install_requires = [
     'arrow>=0.12.1',
+    'attrs>=17.4.0',
```
lnielsen: What is this needed for?

invenio_records_rest/config.py
drjova: 👍 done

setup.py
drjova: There is a compatibility issue between ``attrs`` and ``jsonschema`` with the lowest requirements in travis 
```python
class TypeChecker(object):
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/jsonschema/_types.py", line 64, in TypeChecker
    _type_checkers = attr.ib(default=pmap(), converter=pmap)
TypeError: attrib() got an unexpected keyword argument 'converter'
```
```
jsonschema 3.0.1 has requirement attrs>=17.4.0, but you'll have attrs 17.3.0 which is incompatible.
```

https://travis-ci.org/inveniosoftware/invenio-records-rest/jobs/514094978
will investigate more

drjova:
@slint shall I add a release commit as well?

slint:
Fine by me :) I guess patch version bump (i.e. `v1.4.1`) should be enough

drjova:
@slint done :)

drjova:
@slint ping ☺️

Issue #243, opened 2019-02-06
The sqlalchemy extension was not registered to the current application


jacgeo:
How to get the individual records in invenio-rest-records?
When try to access the individual record through the link `self` getting an error
``` The sqlalchemy extension was not registered to the current application. Please make sure to call init_app() first.```

lnielsen:
Can you provide a bit more context. I.e. did use the cookiecutter-invenio-instance etc....

PR #253, opened 2019-05-12
views: small fix on pass_record raise block

egabancho:
The error matches with the title of #243 `local variable 'pid' referenced before assignment` but the description of that issue seems to be for something different so I didn't add the closes ...

Also, I made the PR on `main-1.2` because https://invenio.readthedocs.io/en/latest/releases/maintenance-policy.html#policy

> A minor release A.B (e.g. v3.0) is supported with bug and security fixes (via patch releases) until the release of A.B+2 (e.g. v3.2) and minimum one year.

Let me know if you want to change this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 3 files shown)


egabancho:
@diegodelemos Thanks!

PR #260, opened 2019-07-15
CSV serialize

vladbmihaescu:
Adding class for serializing records into CSV format.

### DIFF
--- a/invenio_records_rest/serializers/csv.py
+++ b/invenio_records_rest/serializers/csv.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2016-2018 CERN.
+# Copyright (C) 2016-2019 CERN.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -11,10 +11,25 @@
 from __future__ import absolute_import, print_function
 
 import csv
-import tempfile
 
 from .base import PreprocessorMixin, SerializerMixinInterface
 from .marshmallow import MarshmallowMixin
+
+
+class Line(object):
+    """Object that implements an interface the csv writer accepts."""
+
+    def __init__(self):
+        """Initialize."""
+        self._line = None
+
+    def write(self, line):
+        """Write a line."""
+        self._line = line
+
+    def read(self):
+        """Read a line."""
+        return self._line
 
 
 class CSVSerializer(SerializerMixinInterface, MarshmallowMixin,
@@ -28,12 +43,20 @@ class CSVSerializer(SerializerMixinInterface, MarshmallowMixin,
     def __init__(self, *args, **kwargs):
         """Initialize CSVSerializer.
 
-        :param csv_excluded_fields: list of complete paths of the fields that
+        :param csv_excluded_fields: list of paths of the fields that
                                     should be excluded from the final output
+        :param csv_included_fields: list of paths of the only fields that
+                                    should be included in the final output
         :param header_separator: separator that should be used when flattening
                                  nested dictionary keys
         """
         self.csv_excluded_fields = kwargs.pop("csv_excluded_fields", [])
+        self.csv_included_fields = kwargs.pop("csv_included_fields", [])
+
+        if self.csv_excluded_fields and self.csv_included_fields:
+            raise ValueError(
+                "Please provide only fields to either include or exclude")
+
         self.header_separator = kwargs.pop("header_separator", "_")
         super(CSVSerializer, self).__init__(*args, **kwargs)
 
@@ -47,7 +70,7 @@ class CSVSerializer(SerializerMixinInterface, MarshmallowMixin,
         record = self.process_dict(
             self.transform_record(pid, record, links_factory))
 
-        return self.format_csv([record])
+        return self._format_csv([record])
 
     def serialize_search(self, pid_fetcher, search_result, links=None,
                          item_links_factory=None):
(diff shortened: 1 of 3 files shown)


invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+            records.append(self.process_dict(processed_hit, filter_keys))
+
+        return self.format_csv(records)
+
+    def format_csv(self, records):
+        """Format list of flattened dictionaries into CSV format."""
+        headers = sorted(set([key for record in records for key in record]))
+
+        rows = [["\"" + str(record[key]) + "\"" if key in record else "" for key in headers] for
```
ntarocco: I think we can try to optimize this, maybe we can quickly discuss this together. IMHO you could try to build your `set` of keys for the headers while iterating and building the rows.

invenio_records_rest/config.py
```diff
@@ -366,3 +366,6 @@ def deleted_pid_error_handler(error):
 
 RECORDS_REST_DEFAULT_RESULTS_SIZE = 10
 """Default search results size."""
+
+
+FILTER_KEYS = {}
```
ntarocco: We need to find a better way for this config.

tests/test_serializer_csv.py
```diff
@@ -0,0 +1,154 @@
...
+    titles = fields.Raw(attribute='metadata.titles')
+    booklinks = fields.Raw(attribute='metadata.booklinks')
+    circulation = fields.Raw(attribute='metadata.circulation')
+    books = fields.Raw(attribute='metadata.books')
+    document_types = fields.Raw(attribute='metadata.document_types')
+    languages = fields.Raw(attribute='metadata.languages')
+    keywords = fields.Raw(attribute='metadata.keywords')
+    series_objs = fields.Raw(attribute='metadata.series_objs')
+
+    record_1 = {'titles': {"pid": 51, "other": 'A'},
```
ntarocco: can you please add one or multiple values to test corner cases, for example:
* something like `'Jane " Doe'`
* something like `"Jane ' Doe"`
* something like `'Jane "aka" Doe'`
* something like `'François Doe'`
* any other you can think of

invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,118 @@
...
+        return csv_string
+
+    def process_dict(self, dictionary, filter_keys):
+        """Returns a flattened dictionary as a string."""
+        flattened = {}
+
+        self.flatten(dictionary, flattened, filter_keys)
+        return flattened
+
+    def flatten(self, elem, flattened, filter_keys, parent_key='',
```
ntarocco: I am not an extreme fan of not pure functions, in particular with recursion, maybe we can see together if we can find a better way!

ntarocco:
Good job! 👍 Please have a look to Travis too that is failing!

invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,109 @@
...
+                    del flattened[key]
+
+        return flattened
+
+    def flatten(self, elem, flattened, parent_key=''):
+        """Flattens nested dictionaries."""
+        if type(elem) is dict:
+            for key in elem:
+                self.flatten(elem[key], flattened, parent_key + key + '_')
+        elif type(elem) is list:
```
egabancho: I've always found more pythonic something like `isinstance(elem, MutableSequence)` but I guess it's just personal preference ☺️

invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,109 @@
...
+            index = 0
+            for item in elem:
```
egabancho: `for index, item in enumerate(elem)` ?

ntarocco:
@vladbmihaescu @egabancho @slint optimized the code with in memory buffer for the CSV output and return a generator. It should be more efficient with large amount of results. Added tests.
I have tested with a REST client and it looks working fine.

invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def _format_csv(self, records):
+        """Return the list of records as a CSV string."""
+        # build a unique list of all records keys as CSV headers
+        headers = set()
+        for rec in records:
+            headers.update(rec.keys())
+
+        # write the CSV output in memory
+        output = StringIO()
```
slint: TBH, I prefer the second solution from [this SO thread](https://stackoverflow.com/a/32609533) (the one with the `Line` class), it looks a bit higher-level (i.e. no `.seek` and `.truncate` calls)

ntarocco: You are right, this is much better! Will change that! 👍

invenio_records_rest/serializers/csv.py
```diff
@@ -0,0 +1,132 @@
...
+    """CSV serializer for records.
+
+    Note: This serializer is not suitable for serializing large number of
+    records.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialize CSVSerializer.
+
+        :param csv_excluded_fields: list of complete paths of the fields that
```
lnielsen: I must admit I find it more intuitive to specify what to include instead of what to exclude, but that's my personal opinion. I.e. either you include all if the user didn't specify anything, or you include just what the user specfied.

lnielsen:
Please also add tests with unicode characters (this is usually how I break the Python csv module)

lnielsen:
Btw, please also ensure you add it to the documentation:
https://github.com/inveniosoftware/invenio-records-rest/blob/master/docs/api.rst

ntarocco:
> Please also add tests with unicode characters (this is usually how I break the Python csv module) and check the import into e.g. Excel or other spreadsheet to ensure it loads nicely by default.

actually my fault about this, I asked him to do it previously, he did it and then I pushed again removing that by mistake :) @vladbmihaescu sorry!

PR #264, opened 2019-07-29
views: Add support for elements required for infinite scroll.

### LINKED ISSUE #258: views: support "from" parameter in addition to "page" for use with infinite scroll
Trying to implement an infinite scroll with the Invenio REST API is somewhat complicated from the fact that you need to provide ``page`` + ``size`` parameters. Would be easier if it was ``from`` and ``size``. Also, it would be useful to be able to turn off aggregations for all but the first request (when in an infinite scroll). Via Cottage Labs

### LINKED ISSUE #261: views: support for turning off aggregations to support infinite scroll
Related to #258 

When you implement infinite scroll you don't need to have the aggregations recalculated on every update, thus it would be nice with a feature that could switch it off, if not needed.

pazembrz:
Added support for from_record parameter.
Added support for facets parameter which allows to turn on specified
facets or to turn off all of them.

closes #258
closes #261

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,13 @@ extras_require = {
         'pyld>=0.7.1',
     ],
     'tests': tests_require,
+    # Marshmallow v3.x drops support for Python 2
+    ':python_version<"3.0.0"': [
+        'marshmallow>=2.15.2,<3.0.0',
+    ],
+    ':python_version>="3.0.0"': [
+        'marshmallow>=2.15.2',
+    ],
 }
 
 extras_require['all'] = []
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -61,6 +61,13 @@
         'pyld>=0.7.1',
     ],
     'tests': tests_require,
+    # Marshmallow v3.x drops support for Python 2
+    ':python_version<"3.0.0"': [
+        'marshmallow>=2.15.2,<3.0.0',
```
slint: @kprzerwa I noticed that in `invenio-rest` you've specified [`marshmallow v2.15.1` as lowest](https://github.com/inveniosoftware/invenio-rest/commit/e4acef161f0f4b7291d88d262924089b67d30680#diff-2eeaed663bd0d25b7e608891384b7298R40). Since `webargs 5.x` requires [`marshmallow v2.15.2` as lowest](https://github.com/marshmallow-code/webargs/blob/2cf69ec85f80f97eb61848fc1162596e110096b5/setup.py#L6), I'm curious if you had a specific reason in mind?

lnielsen: @slint will you fix it in invenio-rest?

slint: :+1: I checked with @kprzerwa and she said it was like that there. I'm making a PR

slint: Actually, `webargs` is not used in `invenio-rest`, not sure if it should be a dependency...

Issue #250, opened 2019-04-10
browser caches all responses of same url regardless `Content-Type`


harunurhan:
When you rely on `Content-Type` and different serializers of same record to serve different formats under same url, you might have caching issues on most of the browsers. Because invenio sets `Last-Modified` and `Etag` based on only record's version. 

So it's likely that browser will reply with a different format to requests that has the same url. For example `application/json` instead of `application/x-bibtex`, because of same `Etag` and `Last-Modified`

Quick workaround would be remove `Last-Modified` and put `Etag` that includes both `version` and `Content-Type`.

equadon:
## After discussion with Lars

### Current implementation

https://github.com/inveniosoftware/invenio-records-rest/pull/273

We cannot use an `ETag` which uses both the revision id and content-type with the current implementation since `ETag` are used both as a caching mechanism and for concurrency control. For caching it works:
1. Create record
2. GET xml
3. GET json

will get you the right content and caching works.

However, if you do:
1. Create record
2. GET xml
3. GET json
3. PUT and update record

This probably will fail (please verify) and we need to find a solution that separates these concerns.

### Suggestions on where to go next

- Investigate if strong vs. weak ETags could be a solution
- Find a way to separate ETag usage for caching and concurrency control
- Ensure that `check_etag` works correctly with weak and strong ETags
- Now strong ETags are used by default - investigate if weak should be used by default

PR #273, opened 2019-08-02
REVIEW NEEDED [WIP] global: add vary header to the response

kpsherva:
FOR THE REVIEWERS: this is a proposal of the solution, it will be applied not only for the get method but please first share your thoughts

( addresses #250)

**Update:**
The main issue seemed to be that the vary header had no effect if the `etag` was unchanged so the new WIP version has the `etag` include not only the `record.revision_id` but also the mimetype. It's still unclear which effect the `Vary` header has but without it sometimes it doesn't recognize if a change is made in the `Accept` header so it seems to be needed.

Requires https://github.com/inveniosoftware/invenio-rest/pull/100

**PAST NOTES ON THE ISSUE:**

THIS IS not working fully - below my investigation
How to test:
1. Create clean instance
2. Install latest invenio 3.2.0a5
3. install invenio-records-rest from this PR
3.1. on your clean instance create custom serializer of the record f.e. application/x-custom mimetype - make sure it returns also this content type as a response example: https://github.com/kprzerwa/test-headers-instance/blob/master/my_site/records/config.py

You can use firefox for debugging - it has neat way of editing headers in the dev tools

4. post a new record
5. go to `/records/1`
6. response should have `Vary: Content-Type, Accept` header - check this PR
7. Hit again the same endpoint - the response should be cached. But it will not have the Vary header anymore! <--- I think this is where the problem is located
8. change the Accept header to your custom `application/x-custom` in the request and request again the endpoint. You will get the same json response, cached, while you should get 200 status code response with x-custom content type. 
What is probably the problem here: The cached response is not adding the Vary header so the browser does not know that it should check the Accept value every time, and relies only on the combination of `ETag` (which is version of the record) passed to `If-None-Match` header and `If-Modified-Since` header which won't change in this case.

### DIFF
--- a/invenio_records_rest/serializers/response.py
+++ b/invenio_records_rest/serializers/response.py
@@ -30,6 +30,7 @@ def record_responsify(serializer, mimetype):
         response.status_code = code
         response.set_etag(str(record.revision_id))
         response.last_modified = record.updated
+        response.vary = 'Content-Type, Accept'
         if headers is not None:
             response.headers.extend(headers)
 
--- a/invenio_records_rest/serializers/response.py
+++ b/invenio_records_rest/serializers/response.py
@@ -14,6 +14,7 @@ Responsible for creating a HTTP response given the output of a serializer.
 from __future__ import absolute_import, print_function
 
 from flask import current_app
+from invenio_rest.views import create_etag
 
 
 def record_responsify(serializer, mimetype):
@@ -28,7 +29,7 @@ def record_responsify(serializer, mimetype):
             serializer.serialize(pid, record, links_factory=links_factory),
             mimetype=mimetype)
         response.status_code = code
-        response.set_etag(str(record.revision_id))
+        response.set_etag(create_etag(record.revision_id, mimetype))
         response.last_modified = record.updated
         response.vary = 'Content-Type, Accept'
         if headers is not None:
@@ -39,6 +40,7 @@ def record_responsify(serializer, mimetype):
 
         return response
 
+    view.mimetype = mimetype
     return view
 
 
--- a/tests/test_views_item_get.py
+++ b/tests/test_views_item_get.py
@@ -10,6 +10,7 @@
 
 from __future__ import absolute_import, print_function
 
+import pytest
 from flask import url_for
 from helpers import get_json, record_url, to_relative_url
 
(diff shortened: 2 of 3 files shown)


invenio_records_rest/serializers/response.py
```diff
@@ -30,6 +30,7 @@ def view(pid, record, code=200, headers=None, links_factory=None):
         response.status_code = code
         response.set_etag(str(record.revision_id))
         response.last_modified = record.updated
+        response.vary = 'Content-Type, Accept'
```
harunurhan: Not sure if anybody is relying on them, but I think it's better to use other `Accept-` headers too, especially `Accept-Language`

topless: I think since you provide `Accept` the `Content-Type` is redundant.
I would go with  `response.vary = 'Accept, Accept-Encoding, Accept-Language'`

tests/test_views_item_get.py
```diff
@@ -75,3 +76,49 @@ def test_item_get_invalid_mimetype(app, test_records):
         # Check that GET with non accepted format will return 406
         res = client.get(record_url(pid), headers=[('Accept', 'video/mp4')])
         assert res.status_code == 406
+
+
+@pytest.mark.skip
```
topless: We have to enable it

tests/test_views_item_get.py
```diff
@@ -75,3 +76,49 @@ def test_item_get_invalid_mimetype(app, test_records):
...
+        assert res.status_code == 304
+
+        res = client.get(record_url(pid),
+                         headers=[('Accept', 'application/x-custom'),
+                                  ('If-Match', etag_one),
+                                  ('If-Modified-Since',
+                                   last_modified),
+                                  ])
+        # this reveals the problem, should be 200,
+        # can be tested with a browser
```
topless: clean up

invenio_records_rest/serializers/response.py
kprzerwa: once this PR gets approved please use the headers like @topless is suggesting

harunurhan:
As far as I have seen, `Vary` doesn't seem to fix anything. Browsers would still send `If-Match: <Etag>` for requests with `Accept: <different content-type than the request where we got the Etag in If-Match>`, then server would return `304`and cause the same bug.

PR #279, opened 2019-09-16
views: fix wrong `size` param upper limit

ntarocco:
* Since the `size` url param is internally passed to the ES
  query as Python list[start:size], its upper limit value should be
  equal to `max_result_window`

### DIFF
--- a/tests/test_views_search.py
+++ b/tests/test_views_search.py
@@ -51,8 +51,16 @@ def test_page_size(app, indexed_records, search_url):
         res = client.get(search_url, query_string=dict(page=1, size=10))
         assert_hits_len(res, len(indexed_records))
 
-        # Exceed max result window
+        # Last page
         res = client.get(search_url, query_string=dict(page=100, size=100))
+        assert_hits_len(res, 0)
+
+        # Exceed max result window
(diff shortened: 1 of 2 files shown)


tests/test_views_search.py
```diff
@@ -318,14 +358,14 @@ def test_from_parameter_invalid_pagination(
         assert errors == {('from', 'Must be at least 1.'), } or \
             errors == {('from', 'Must be greater than or equal to 1.'), }
 
-        res = client.get(search_url, query_string={'size': 1, 'from': 9999})
+        res = client.get(search_url, query_string={'size': 1, 'from': 10000})
```
slint: Not a blocker for me, but I have a slight doubt about this one in terms of behavior. If I request `size: 1, from: 1` I would expect to get 1 result back (i.e. result no. 1 out of 10.000). Keep in mind that it's not possible to do `size: 1, from: 0` since `from` has to be greater that 0 (there's a test above for that). If I wanted to get the results no. 10.000 I would in the same fashion do `size: 1, from: 10000`, which would return only one record, the last one.

ntarocco: This is just an issue with the demo data not having 10K results. I have updated the tests hopefully in a more understandable way.

PR #288, opened 2020-01-31
marshmallow: support load_from/data_key in StrictKeysMixin

ppanero:
**Issue**: The StrictKeysMixin does not support schemas that use `load_from` (Marshmallow 2.x) or `data_key` (Marshmallow 3.x).

**Why?**: Because it matches the *original_data* against a generated list of fields (i.e. from the schema). Therefore, if you are loading from a different field name it will fail. Example:

``` python

class TestSchema(StrictKeysMixin):
    """Test schema."""
    _und_field = fields.Str(data_key='und_field')

test = {"und_field": "Underscore gotten!"}
```

The current `StrictKeysMixin` would look for `_und_field` but the original data has `und_field`. Therefore, failing.

**Solution**: Check for the value of `load_from` or `data_key` in the `Field` object, it is not `None` use that instead of the Schema name.

**Bonus points**: The list of fields was calculated for every step of the `for key in original_data` iteration. Now it is calculated only once ;) !

**Note**: The code is supposed to be Marshmallow 2.x "compatible". Fails in the case of `custom_fields`, I am not super sure why. However, Do we still need backward-compatibility with Python 2 / Marshmallow 2? Yes, right?

### DIFF
--- a/invenio_records_rest/schemas/json.py
+++ b/invenio_records_rest/schemas/json.py
@@ -22,6 +22,18 @@ from invenio_records_rest.schemas.fields import PersistentIdentifier
 class StrictKeysMixin(Schema):
     """Ensure only valid keys exists."""
 
+    def _calculate_field_list(self):
+        _fields = []
+        if marshmallow_version[0] < 3:
+            for name, field in self.fields.iteritems():
+                aux = self.fields[name].attribute or field
+                _fields.append(aux.load_from or name)
+        else:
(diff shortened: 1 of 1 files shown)


ppanero:
Marshmallow 3.x supports this by default. This would fix Marshmallow 2.x but since no one has hit the issue/we should deprecate at some point I am closing the PR.

PR #301, opened 2020-05-11
global: fix for classifiers

### LINKED ISSUE #300: Fix classifiers in setup.py
Currently stating python2 to 3.5 when supported is 3.6 and 3.7

topless:
- closes #300

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.7.2 (to-be-determined)
+
+- Fixes classifiers to reflect Python verisons
+
 Version 1.7.1 (released 2020-05-07)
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.7.2 (to-be-determined)
+
+- Fixes classifiers to reflect Python verisons
```
ppanero, line 13: comment:
Do we need to trigger a release? 

- I agree with doing the changelog step by step, so I agree with this.
- I wonder, since the classifiers do not actually affect where it can be installed, they are mere text... Maybe we can until we have actual content/features?

topless: I wouldn't trigger a release, I just put them in the changelog so they pile up with the rest of the changes that will come in the release.

PR #302, opened 2020-07-22
global: add Cache-Control header

ntarocco:
- adds Cache-Control:'no-cache' header to 200 responses to
  ensure that browsers will not cache responses client side
- fixes isort to support v5


ntarocco:
Do we want to depend on `invenio-rest` specific min version, after the release of https://github.com/inveniosoftware/invenio-rest/pull/114? Personally, I would say no...

lnielsen (approved):

PR #304, opened 2020-08-21
[DO NOT MERGE] This PR is for demo only

ppanero:
This PR is to display the changes needed in order to use the new `docker-services`:

- Adapt `.travis.yml`. In the end is more of a cleanup, since is merely removing services from travis itself and then adapting a few env vars to provide service's versions.
- Adapt `run-tests.sh`. Mainly requires to add a call to start and one to stop the services.

The rest of the changes are there to avoid things like `isort` issues to block the run of pytest.


.travis.yml
```diff
@@ -64,12 +53,10 @@ before_install:
 install:
   - "travis_retry pip install -r .travis-${REQUIREMENTS}-requirements.txt"
   - "travis_retry pip install -e .[$EXTRAS]"
+  - "travis_retry pip install git+https://github.com/ppanero/docker-services.git#egg=docker-services"
```
lnielsen, line 56: This would be a PyPI package, right?

run-tests.sh
```diff
@@ -2,13 +2,15 @@
...
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
-pydocstyle invenio_records_rest tests docs && \
-isort -rc -c -df && \
-check-manifest --ignore ".travis-*" && \
-sphinx-build -qnNW docs docs/_build/html && \
-python setup.py test
+docker-services-cli services up es postgresql redis
```
lnielsen, line 10: This would require docker-services-cli to be installed right? Shouldn't this be a test dependency in ``setup.py`` then, or you plan to install it via ``pytest-invenio``?

.travis.yml
```diff
@@ -20,41 +14,36 @@ language: python
...
+    - REQUIREMENTS=release EXTRAS=all,elasticsearch7 ES_VERSION=ES_7_LATEST
+    - REQUIREMENTS=devel EXTRAS=all,elasticsearch7 ES_VERSION=ES_7_LATEST
 
 python:
   - "3.6"
   - "3.7"
   - "3.8"
 
 before_install:
+  - sudo rm /usr/local/bin/docker-compose
```
lnielsen, line 42: Comment: Could these lines be reduced to a oneliner via e.g. automation with docker-services-cli?

run-tests.sh
lnielsen: Basically, what I'm looking for is that the following would still work:

```
git clone ...
cd invenio-<module>
pip install -e .[all]
./run-tests.sh
```

lnielsen: and something like this for running locally with something else:

```
env ES_VERSION=ES5_LATEST ./run-tests.sh
```

.travis.yml
diegodelemos: Yes

run-tests.sh
ppanero: Should come with pytest-invenio or be a test dependency yes. My thought was with `pytest-invenio`

.travis.yml
ppanero: yes, this is here until we actually release it. just for demo

.travis.yml
ppanero: this is needed for the travis vm... we could potentially run this after the install step (line 56). However, I don't think this pakcage should change "binaries of the computer". IMO it falls in the same category than `pre-requisites` in RDM... *If you want to use this, you need this that and that other min versions*

PR #311, opened 2021-02-08
es: boost fields in search factory

KonstantinaStoikou:
Boost fields defined in `boosted_fields` attribute of a `RecordsSearch` class (if the attribute exists).


ntarocco:
@KonstantinaStoikou thanks a lot for this PR! Would it be possible to add a test to show the boosting?
Or too hard? I am thinking to have a couple of demo records where we boost the title for example...

We also need to see this issue with the `*`.

invenio_records_rest/query.py
```diff
@@ -25,8 +25,9 @@ def default_search_factory(self, search, query_parser=None):
     """
     def _default_parser(qstr=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
+        fields = getattr(search, "boosted_fields", []) + ["*"]
```
ntarocco, line 28: Given the issue with the `fields`/`lenient`, what about if we rewrite this? (and move it inside the if)
```
if qstr:
    boosted = getattr(search, "boosted_fields", [])
    extra_params = {}
    if boosted:
        extra_params["fields"] = boosted + ["*"]
        extra_params["lenient"] = True  # when listing fields that are not string (number, date) ignore formatting errors
    return Q("query_string", query=qstr, **extra_params) 
```

We have also to document this, otherwise the `boosted_fields` attr is very hidden.

tests/conftest.py
```diff
@@ -49,6 +49,8 @@
 class TestSearch(RecordsSearch):
     """Test record search."""
 
+    boosted_fields = ["title^5"]
```
kprzerwa, line 52: could this be a part of RECORDS_REST_CONFIG ? in this way is a bit hidden. ping @ntarocco

PR #322, opened 2023-05-23
add init function

utnapischtim:
- fix: before_app_first_request deprecation
- tests: json error message has been changed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@
 #
 # This file is part of Invenio.
 # Copyright (C) 2020 CERN.
-# Copyright (C) 2022-2023 Graz University of Technology.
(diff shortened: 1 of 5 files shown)


slint (commented):
I think this module is deprecated... In this case, we'll probably archive it or just stop maintaining it.

utnapischtim:
@slint yes it is deprecated, but it is still in use. (example: [invenio-github](https://github.com/inveniosoftware/invenio-github/blob/dc7fa302e5c9fd81cd777e8bfe024ef502661069/setup.cfg#L61), [invenio-app-rdm](https://github.com/inveniosoftware/invenio-app-rdm/blob/23295ca7a696c6ecaaca012fe62b1588fdbd867f/setup.cfg#L57)) if we don't maintain it, we move to flask>=2.3.

PR #327, opened 2023-11-16
facets: New parameter, RECORDS_REST_FACETS_FILTER, to filter the face…

psaiz:
…ts based on a category based on all the other categories

:heart: Thank you for your contribution!

### Description

This is inspired by the work of @ParthS007  [here](https://github.com/cernopendata/opendata.cern.ch/commit/4cd8bd186d04286aaedbc8a09ccd4938bc96d98d)

The idea is to offer the possibility to apply the filters to the aggregations. As an example:
* Imagine a search with two different categories,  `cat1`  and `cat2`.
* The user selects `cat1=value1`
* The current options are that either that selection is applied to the results, and the aggregation (standard filter), or only to the results (`post_filter`). 
* This PR introduces a new option, where the filter is applied to the results, and to all the other aggregations (in the example, the aggregation of `cat2` will return only the entrie that satisfy `cat1=value1`, but the aggregation for `cat1` returns everything.

### DIFF
--- a/invenio_records_rest/config.py
+++ b/invenio_records_rest/config.py
@@ -344,6 +344,9 @@ The structure of the dictionary is as follows:
     }
 """
 
+RECORDS_REST_FACETS_POST_FILTERS_PROPAGATE = False
+"""Define if the post_filters facets in one category should be applied as filters to all the other categories"""
+
 RECORDS_REST_DEFAULT_CREATE_PERMISSION_FACTORY = deny_all
 """Default create permission factory: reject any request."""
 
--- a/invenio_records_rest/facets.py
+++ b/invenio_records_rest/facets.py
@@ -22,21 +22,33 @@ from invenio_records_rest.utils import make_comma_list_a_list
 
 
 def nested_filter(field, subfield):
-    """Create a nested filter. Similar to the https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/facets/facets.py#L94"""
+    """Create a nested filter.
+
+    Similar to the example from
+    https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/facets/facets.py#L94
+    """
 
     def inner(values):
         top_level = []
         queries = []
         for value in values:
             subvalues = value.split("::")
-            if len(subvalues)>1:
-                queries.append(dsl.Q("bool", must=[dsl.Q("term", **{field: subvalues[0]}), dsl.Q("term", **{subfield: subvalues[1]})]))
+            if len(subvalues) > 1:
+                queries.append(
+                    dsl.Q(
+                        "bool",
+                        must=[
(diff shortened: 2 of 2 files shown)


invenio_records_rest/facets.py
```diff
@@ -138,11 +138,37 @@ def _query_filter(search, urlkwargs, definitions):
 
     return (search, urlkwargs)
 
+def without_nested_subtypes(facet_filters, facet_names):
+    """Remove the nested subtypes from the filter.
 
-def _aggregations(search, definitions):
+    Example: If `CMS` from Experiment type is selected
```
ntarocco, line 144: Very minor: we tend to use examples (or maybe it is only me 🤣) not domain specific, so that it is more clear. In RSK, we are using resource types [here](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/state/selectors/query.test.js#L17).

invenio_records_rest/facets.py
```diff
@@ -138,11 +138,37 @@ def _query_filter(search, urlkwargs, definitions):
...
+    for name in facet_names:
+        new_facet_filters.pop(name)
+    return new_facet_filters
+
+def _aggregations(search, definitions, updated_filters=None, urlkwargs=None):
     """Add aggregations to query."""
     if definitions:
         for name, agg in definitions.items():
+            if name in updated_filters:
+                (facet_filters, _)= _create_filter_dsl(urlkwargs, without_nested_subtypes(updated_filters, [name]))
```
ntarocco, line 158:
```suggestion
                facet_filters, _ = _create_filter_dsl(urlkwargs, without_nested_subtypes(updated_filters, [name]))
```

ntarocco (commented):
Would it be possible to go through the changes IRL?
I feel like we require more comments or to extract more small func to increase clarity :)

ntarocco:
Tests are failing: it might be that you first want to have the changes validated?

invenio_records_rest/config.py
```diff
@@ -344,6 +344,9 @@ def deleted_pid_error_handler(error):
     }
 """
 
+RECORDS_REST_FACETS_FILTER = False
```
ntarocco, line 347:
```suggestion
RECORDS_REST_FACETS_POST_FILTERS_PROPAGATE = False
```

invenio_records_rest/facets.py
```diff
@@ -138,11 +138,37 @@ def _query_filter(search, urlkwargs, definitions):
 
     return (search, urlkwargs)
 
+def without_nested_subtypes(facet_filters, facet_names):
```
ntarocco, line 141: Improve naming

invenio_records_rest/facets.py
psaiz: Done, thanks

ntarocco (approved):

ntarocco (approved):

PR #352, opened 2025-07-18
fix errors

utnapischtim:
- fix: remove datacite schema31 support
- feat: add schema support up to 45


tmorrell (approved):
Wow, I didn’t know these were defined here since they aren’t used in rdm. I don’t think there is a need to update with the newer ones unless someone is using them.

utnapischtim:
perfect, than this is enough :)

PR #365, opened 2026-06-18
fix(views): pass_record swallows StaleDataError as PIDResolveRESTError

jma:
The return f(...) call inside pass_record's try/except SQLAlchemyError
caused any database error raised by the handler (e.g. StaleDataError on
optimistic-locking conflict) to be silently re-raised as PIDResolveRESTError
(500) instead of propagating correctly.

- Move return f(...) outside the try block in pass_record so the except
  only covers PID resolution.
- Handle StaleDataError explicitly in put() with rollback + 409.
- Handle SQLAlchemyError explicitly in delete() with rollback + 500,
  restoring the behaviour previously masked by pass_record.
- Add tests for the stale-data 409 path and for pass_record no longer
  swallowing handler-level SQLAlchemy errors.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

:heart: Thank you for your contribution!

### Description

> Please describe briefly your pull request.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_records_rest/views.py
+++ b/invenio_records_rest/views.py
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2015-2019 CERN.
 # SPDX-FileCopyrightText: 2023-2026 Graz University of Technology.
+# SPDX-FileCopyrightText: 2026 RERO.
 # SPDX-License-Identifier: MIT
 
 """REST API resources."""
@@ -32,6 +33,7 @@ from invenio_search.engine import search as search_engine
 from jsonpatch import JsonPatchException, JsonPointerException
 from jsonschema.exceptions import ValidationError
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.orm.exc import StaleDataError
 from webargs import ValidationError as WebargsValidationError
 from webargs import fields, validate
 from webargs.flaskparser import parser
@@ -44,6 +46,7 @@ from .errors import (
     JSONSchemaValidationError,
     PatchJSONFailureRESTError,
     PIDResolveRESTError,
+    RecordConflictRESTError,
     SearchPaginationRESTError,
     SuggestMissingContextRESTError,
     SuggestNoCompletionsRESTError,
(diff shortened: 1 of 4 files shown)


invenio_records_rest/views.py
```diff
@@ -945,8 +951,12 @@ def put(self, pid, record, **kwargs):
...
         record.clear()
         record.update(data)
-        record.commit()
-        db.session.commit()
+        try:
+            record.commit()
+            db.session.commit()
+        except StaleDataError:
+            db.session.rollback()
+            abort(409)
```
max-moser: For a second, I thought maybe this would better fit into an error handler, but it seems like handling errors in the view functions is a pattern used in other view functions here too; e.g. in `patch()` https://github.com/inveniosoftware/invenio-records-rest/blob/v6.0.1/invenio_records_rest/views.py#L902-L905

max-moser (approved):
Fixes accidental swallowing + re-raising of an exception type, which caused error semantics to be muddled.
Having clearer error semantics is a win in my book.

invenio_records_rest/views.py
jma: Yes you are right, I will use a similar approach than: https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/errors.py.

jma: Done, what do you think?

ntarocco (approved):
LGTM! Thank you for the fix

